PR #3, opened 2015-09-26
Add DynamoDB Document API support for DynamoDB Streams

hansonchar:
In summary, this is a proposed change to provide DynamoDB Document API support for DynamoDB Streams, so that Stream events can be processed at the Document API level, dealing with the high-level artifacts such as Item or PrimaryKey, instead of the low-level artifacts such as maps or AttributeValue.


ahmedammar:
:+1: to pull this

asuresh8:
Support for Dynamodb events already exists

PR #9, opened 2016-03-25
Update to use hard versions of dependencies

ryano144:
The open ended [1.10.5,) causes maven/ivy to go back to the central repo to look for the latest version every time the build runs.  At least in SBT, this forces the latest version of the SDK to download every time it is released, which has caused problems for our build.  It also makes dependency resolution take an unnecessarily long time.  This update chooses the current latest version of the SDK.

### DIFF
--- a/aws-lambda-java-events/pom.xml
+++ b/aws-lambda-java-events/pom.xml
@@ -3,7 +3,7 @@
   
   <groupId>com.amazonaws</groupId>
   <artifactId>aws-lambda-java-events</artifactId>
-  <version>1.1.0</version>
+  <version>1.1.1</version>
   <packaging>jar</packaging>
 
   <name>AWS Lambda Java Events Library</name>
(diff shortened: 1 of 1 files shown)


torgeir:
Second this one, as we've hit several issues with NoClassDefFoundError errors in our lambdas due to the lax version ranges of the `aws-lambda-java-events` pom. While minor and patch releases generally should not break apis this way, these version ranges would also bump major versions which definitely would break apis. 

Temporarily we exclude all dependencies we actually depend on from the aws-lambda-java-events module to lock down version numbers in explicit dependencies instead

```
<dependency>
  <artifactId>aws-lambda-java-events</artifactId>
  <version>${aws.lambda.version}</version>
  <exclusions>
    <exclusion>
      <artifactId>aws-java-sdk-s3</artifactId>
    </exclusion>
    <exclusion>
      <artifactId>aws-java-sdk-sqs</artifactId>
    </exclusion>
  </exclusions>
</dependency>
```

petehayes:
This can be closed as the latest version removes the dynamic version ranges.

bmoffatt:
https://github.com/aws/aws-lambda-java-libs/commit/617af8467230bf78a486316db8d4591f7c3ae75e  https://github.com/aws/aws-lambda-java-libs/commit/65ef84f630638a149fab6504d938d234eef0da4a

Issue #17, opened 2016-10-13
Proxy Lambda input and output format


sihil:
It would be great to have objects that match up with the input and output formats documented at http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-set-up-simple-proxy.html#api-gateway-simple-proxy-for-lambda-input-format

statelessness:
+1

alsanrum:
+1

brackxm:
see https://github.com/awslabs/aws-serverless-java-container/tree/master/aws-serverless-java-container-core/src/main/java/com/amazonaws/serverless/proxy/internal/model

asuresh8:
Added

PR #19, opened 2016-12-16
Input and output for lambda proxy integration

guimeira:
These two classes map the input and output expected by the lambda proxy integration, as documented here: http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html#api-gateway-create-api-as-simple-proxy-for-lambda-test 

This implements the feature requested on issue #17.


asuresh8:
added in java events v2 update

PR #30, opened 2017-11-04
Update version to 2.0.0 to keep semver consistency with version 1.X

twbarber:


twbarber:
Noticed this version was changed to `2.0`, which is inconsistent with the semver `1.X.X` that was published before.

asuresh8:
Yes we caught this inconsistency a while back and needed to do a patch, so the latest revision is v2.0.1. It looks like we missed updating that pom.xml. We are updating some packages in the next few weeks, and when we do we will update the versions in the pom files.

PR #39, opened 2018-01-16
add authorizer context to APIGatewayProxyRequestEvent

oharaandrew314:
The fields in the authorizer object can change depending on what is returned by the authorizer function context.  While `principalId` will always be there, the rest of the fields can vary.  Thus a `Map` must be used.

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyRequestEvent.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyRequestEvent.java
@@ -55,6 +55,8 @@ public class APIGatewayProxyRequestEvent implements Serializable, Cloneable {
 
         private String path;
 
+        private Map<String, String> authorizer;
+
         /**
          * default constructor
          */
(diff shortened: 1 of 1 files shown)


aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyRequestEvent.java
```diff
@@ -267,6 +269,19 @@ public ProxyRequestContext withPath(String path) {
             return this;
         }
 
+        public Map<String, String> getAuthorizer() {
+            return authorizer;
+        }
+
+        public void setAuthorizer(Map<String, String> authorizer) {
```
bmoffatt: From https://github.com/aws/aws-lambda-dotnet/issues/98, `"authorizer"` may contain more than just strings. To be truly generic, it might be that a `APIGatewayAuthorizedProxyEvent<T>` needs to be added.

miere:
@oharaandrew314 
Do you mind to share the reason of you've closed this PR?

avanathan:
I have re-opened this PR as https://github.com/aws/aws-lambda-java-libs/pull/77. Thanks

PR #52, opened 2018-06-29
make SQSEvent.SQSMessage public

### LINKED ISSUE #51: SQSEvent.SQSMessage class and constructor need to be public
The SQSEvent.SQSMessage class needs to be public and so does its constructor. It's private and completely unusable to us right now.

shunyy:
fix #51 

###  Description of changes
make SQSEvent.SQSMessage public


bmoffatt:
Thanks! Will make 2.2.1 available in Maven soon

PR #58, opened 2018-09-20
Added annotations for documentation

raupachz:
Hi there,

we are using AWS Lambda with Scheduled Events and and as API Gateways for quite a while. Couple years actually and we are quite happy with the results. If you deploy Lambda Functions and forget about them it is often not easy to figure out which class in your JAR file matches the Function Name in the AWS Console. 

If you happen to use a Lambda Function as an endpoint for AWS API Gateway it is even more confusing.

We found out that having to little annotations can make a big difference. It is just for documentation so nothing of the functionality gets touched. 

For example:
```
@FunctionName("metric-service")
@Endpoint("/services/metrics/{date}")
public class MetricsHandler implements RequestHandler ...
```

With the annotations it is now much clearer what the MetricsHandler does.

I know it is just a tiny and ridiculous simple addition, but it was very helpful for us, maybe helpful for others.

kind regards
Björn


msailes:
Hi @raupachz,

This is a great way to document your code, I can imagine it being useful in larger workloads.

I don't think it would be best suited to be added to this library. Would your organization consider open sourcing it within a module?

Thanks,

Mark

Issue #65, opened 2018-12-26
Fix "Paramters" typo in "aws-lambda-java-events"


stepio:
Spotted a typo (correct me if I'm wrong) in two of "with*" methods in `aws-lambda-java-events`.

PR #54, opened 2018-07-09
Fix typo: paramters -> parameters.

vmunier:
*Description of changes:*
Fix a typo: paramters -> parameters.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #66, opened 2018-12-26
GH-65 Fix "Paramters" typo

stepio:
*Issue #65*

Fixed a typo in method's naming, also did a search and fixed the same in a couple of comments.
**Warning:** as existing API methods are renamed, it's a breaking change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,10 @@
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
+
+# Maven build
+target/
+
+# IDEA internal
+*.iml
+.idea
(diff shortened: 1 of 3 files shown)


stepio:
Sorry, my bad, this almost duplicates #54, although I found and fixed one more occurrence in comments.

mukhtar:
Thanks for the fix. (de)serialization doesn't seem to be impacted by the typos, however, it would be preferable to keep the existing methods with the typos for backwards compatibility, mark them as deprecated, and add two new methods with the fixes.

stepio:
Thanks for feedback, you're right, it's safer to keep backward compatibility this way. Updated & force-pushed. Also slightly updated `.gitignore` file: added maven build folder and IntelliJ IDEA metadata files and folders.

PR #77, opened 2019-02-18
Adding authorizer field to collect context properties from authorizer

avanathan:
*Issue #, if available:*
Reopening: https://github.com/aws/aws-lambda-java-libs/pull/39

*Description of changes:*
Adding a map to hold context properties from authorizer. cc: @bmoffatt @oharaandrew314 @miere . This PR was originally submitted by @oharaandrew314 and got closed based on a review comment for using `String` key/values. All other collection fields in this class use `Map<String, String>`. Not sure why we need to worry about other types (non `String`) just for this. We could keep this in the short term until we find a long term solution. Libraries for rest of the programming languages like golang & .Net already support this. Thanks for your support.

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyRequestEvent.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyRequestEvent.java
@@ -29,6 +29,8 @@ public class APIGatewayProxyRequestEvent implements Serializable, Cloneable {
 
     private Map<String, String> stageVariables;
 
+    private Map<String, String> authorizer;
+
     private ProxyRequestContext requestContext;
 
     private String body;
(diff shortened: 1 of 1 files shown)


avanathan:
Let me know if we need to keep non-String types specifically for Authorizer context. I'll change it to `Map<String, Object>`

JustDerb:
+1 to trying to get this added. For people trying to use this library for Cognito User Pool integration (or other types of authorization), it's a non-starter.

avanathan:
@bmoffatt could you please take a look & share your thoughts? Thanks

bmoffatt:
@avanathan `Map<String, Object>` will match what we've got for [dotnet](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayCustomAuthorizerContext.cs#L12) and [go](https://github.com/aws/aws-lambda-go/blob/master/events/apigw.go#L39)

avanathan:
@bmoffatt I've switched it to `Map<String, Object>`

bmoffatt:
Thanks!

avanathan:
@bmoffatt When do we usually cut a release? `2.2.6` ? Is there a way I could do this/help?

PR #92, opened 2019-07-23
Adding support for APIGatewayV2 (Web Sockets)

TimGustafson:
*Issue #, if available:*
N/A

*Description of changes:*
Added support for APIGatewayV2 events.


bmoffatt:
Can you link a reference to the relevant AWS documentation and sample data?

TimGustafson:
The basic AWS documentation about WebSocket APIs is available at:

https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api.html

Documentation specific to the Lambda Proxy integration can be found at:

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html

Is that what you're looking for?

bmoffatt:
Going to use https://github.com/aws/aws-lambda-go/pull/159 as a sample data reference

bmoffatt:
Having 'V2' in the name may be confusing for future readers. Putting Websocket in the type name will be clearer

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2ProxyResponseEvent.java
```diff
@@ -0,0 +1,99 @@
+package com.amazonaws.services.lambda.runtime.events;
+
+import java.io.Serializable;
+import java.util.Objects;
+
+/**
+ * @author Tim Gustafson <tjg@amazon.com>
+ */
+public class APIGatewayV2ProxyResponseEvent implements Serializable, Cloneable {
```
bmoffatt: Suggest changing names to be like `APIGatewayWebSocketProxy....`

TimGustafson: I was following the naming convention of the product.  All the SDKs refer to the product as "APIGatewayV2".  I agree, it's confusing, but it maps to the actual product implementation.

For example:

https://docs.aws.amazon.com/cli/latest/reference/apigatewayv2/index.html

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-api.html

https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/apigatewayv2/package-summary.html

mukhtar:
Can you also add the remaining fields to `APIGatewayV2ProxyRequestEvent`? See aws/aws-lambda-go#159.

TimGustafson:
I've added the static declaration for the inner classes, and also added the missing properties.

mukhtar:
I'm facing an issue with the `authorizer` field which is an object in the sample from the aws-lambda-go repo. It seems it can be either a string or an object according to the aws-lambda-go pr. Any reason you selected `String`?

TimGustafson:
I have changed the authorizer field to be a Map<String,Object> so that it will take just about anything that comes in a map format.  I've tested this against an APIGatewayV2 deployment that have an authorizer configured, and one that does not.  It appears to work in both cases.

In investigating this, I realized that we also don't have classes for APIGatewayV2Authorizer(Request|Response) objects and started creating these to make authorizers easier to author in Java, but ran into a problem where Java serialization lower-cases the first letter of each property by default, but the policyDocument returned by the custom authorizer needs to have "Effect", "Action" and "Resources" all begin with a capital letter.  I'm going to spend a bit more time to see if I can get that to work, as I believe that will provide a superior experience for users, and then I'll also supply some demo code (including a CloudFormation template) to demonstrate how to use these, as doing so right now is non-trivial, and required me to spend quite a bit of time doing trial-and-error until I got it just right.

If I do create those APIGatewayV2Authorizer(Request|Response) classes, should I include them in this fork, or should we close this (since technically they're not required for the APIGatewayV2Proxy(Request|Response)Event objects) and open a new fork for the authorizer classes?

mukhtar:
Thanks! I think it would be great to get the APIGatewayV2Proxy(Request|Response) merged in so if we remove APIGatewayV2Authorizer(Request|Response) we're good to go. I tested this too and LGTM.

TimGustafson:
I have removed the custom authorizer classes.  (Sorry, I didn't realize I had added them to the master branch!)

adamnfish:
Is there any reason the new `APIGatewayV2ProxyResponseEvent` class lacks `withX` methods? The non-v2 version does include `withX` methods and I think that's the common pattern for the AWS SDKs?

Here's an example of an existing with method on the original APIGatewayProxyResponseEvent class:
https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyResponseEvent.java#L44

Thanks very much this contribution, this will be a great help. Any idea when it will be available in a public release?

PR #123, opened 2020-05-05
Add APIGateway event for http payload version 2.0

bmoffatt:
*Description of changes:*

* aws docs reference: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
* equivalent types in other runtimes
  * [aws-lambda-dotnet](https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.APIGatewayEvents/APIGatewayHttpApiV2ProxyRequest.cs)
  * [aws-lambda-go](https://github.com/aws/aws-lambda-go/blob/master/events/apigw.go#L45-L105)

This change also introduces a build-time dependency on lombok for code-generation.

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2HTTPEvent.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2HTTPEvent.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+ * the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0
+ *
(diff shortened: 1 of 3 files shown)


aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2HTTPEvent.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+import org.joda.time.DateTime;
+
+@EqualsAndHashCode
+@Getter
+@NoArgsConstructor
+@Setter
```
carlzogh, line 29: I'd suggest using `@NoArgsConstructor` with `@Data` (https://projectlombok.org/features/Data) as that includes `@EqualsAndHashCode`, `@Getter`, `@Setter`, `@RequiredArgsConstructor` and `@ToString`.

bmoffatt: Much simpler! Will do

PR #125, opened 2020-05-05
Add APIGatewayV2WebSocketEvent/Response as renames of APIGatewayV2Proxy* classes

bmoffatt:
*Description of changes:*

There's now a few things that are "APIGatewayV2" resources. This rename should help websocket users find the type they need.


aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2ProxyRequestEvent.java
```diff
@@ -1,726 +1,4 @@
...
-
-    sb.append("isBase64Encoded: ").append(isBase64Encoded).append(",");
-
-    sb.append("}");
-
-    return sb.toString();
-  }
-
-}
+@Deprecated()
```
carlzogh, line 3: Maybe a message in the `@Deprecated` tag could be helpful in letting users know why the class is deprecated and what they use instead?

Issue #104, opened 2020-01-03
Please add Destinations Invocation Records


mikebroberts:
Per https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html#invocation-async-destinations

Thanks!

Mike

msailes:
Hi Mike, could you tell us a bit more about what you'd like to see?

mikebroberts:
Going back a little now so my memory is fuzzy ... but I think what I'd like is to be able to create Java Lambda functions that can act as Destinations. And therefore "AWS Lambda Java Events" updated to support this scenario. IIRC I tried doing this manually with my own Jackson objects and hit some problems.

msailes:
Thanks for taking the time to explain it, I'll have a look into it.

carlzogh:
Hey @keinproblem,

`LambdaDestinationEvent ` has been added to the `aws-lambda-java-events` library and released as part of version `3.3.0` (ref: https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/RELEASE.CHANGELOG.md).

It is now available to download from Maven https://search.maven.org/artifact/com.amazonaws/aws-lambda-java-events/3.3.0/jar

Thanks for bringing this up!

PR #139, opened 2020-05-26
initial Lambda destination event

msailes:
*Issue #, if available:*
https://github.com/aws/aws-lambda-java-libs/issues/104

*Description of changes:*
Event to represent the Lambda destination event.

Maybe this should be called something else. I'm open to suggestions.

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance with
+ * the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package com.amazonaws.services.lambda.runtime.events;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import org.joda.time.DateTime;
+
+import java.io.Serializable;
+import java.util.Map;
+
+/**
+ * Class to represent an invocation record for a Lambda event.
+ *
+ * @see <a href="https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html>Asynchronous invocation</a>
+ *
+ * @author msailes <msailes@amazon.co.uk>
+ */
+
+@Data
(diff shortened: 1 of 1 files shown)


aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
```diff
@@ -0,0 +1,62 @@
...
+        private String requestId;
+        private String functionArn;
+        private String condition;
+        private int approximateInvokeCount;
+    }
+
+    @NoArgsConstructor
+    @Data
+    public static class ResponseContext implements Serializable, Cloneable {
+        private String statusCode;
```
andreac30, line 52: Looking at the AWS doc for the LambdaDestinationEvent I think statusCode should be an int type.

msailes: Fixed.

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
```diff
@@ -0,0 +1,62 @@
...
+ * http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package com.amazonaws.services.lambda.runtime.events;
+
+import lombok.Data;
+import lombok.NoArgsConstructor;
```
andreac30, line 16: As the other events, I think also here the `NoArgsConstructor` is superfluous.

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
```diff
@@ -0,0 +1,62 @@
...
+        private int statusCode;
+        private String executedVersion;
+        private String functionError;
+    }
+
+    @NoArgsConstructor
+    @Data
+    public static class ResponsePayload implements Serializable, Cloneable {
+        private String errorMessage;
+    }
```
andreac30, line 61: Looking at [aws doc for lambda destination event](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-destinations/) it seems that `ResponsePayload` and `ResponseContext` can have different fields. I would suggest using Map<String, Object> as the `requestPayload` above, what do you think?

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
msailes: fixed

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
msailes: fixed

msailes:
I've just tested this on my account.

PR #149, opened 2020-07-14
Create explicitly getter and setter for the isBase64Encoded variable

galquezar:
*Description of changes:*
Create explicitly getter and setter for the isBase64Encoded because Jackson remove the prefix is from the boolean variables when serializing. ALB expects a key-value isBase64Encoded not base64Encoded in the response JSON document. Thus, the JSON serialization doesn't work properly for Base64 bodies. It is also coherent with the APIGatewayProxy* ones.


galquezar:
After discussing the issue with @carlzogh , we agreed to change the lombok behavior instead of adding explicit getter and setter.

PR #150, opened 2020-07-23
Fixes Base64 enconding for ALB and API Gateway HTTP events

galquezar:
*Issue*
Jackson uses the getters/setters names to name the JSON keys when serializing/deserializing, and Lombok automatically generate getters/setters. A class using the boolean variable IsBase64Encoded would have IsBase64Encoded/SetBase64Encoded methods. Jackson would convert them to base64Encoded key instead of isBase64Encoded. That is incompatible with the ALB and API Gateway events JSON specification. By using lombok.getter.noIsPrefix = true, we would have GetIsBase64Encoded/SetIsBase64Encoded methods, and Jackson would convert them to the expected key isBase64Encoded.

*Description of changes:*
Adds lombok.config with lombok.getter.noIsPrefix = true to change the serialization behavior with the isBase64Encoded key.


carlzogh:
Can validate that this works - thanks! A suggestion would be to move the `lombok.config` file further up in the tree so it doesn't live alongside the events

galquezar:
Sure, moved to src/main/java

Issue #109, opened 2020-04-07
Any plan to add EventBridge event?


alicanguclukol:
Hi there.

I'm using a lambda function to process events from an event bus.

I'm not able to find out which is the runtime event class to access these events.
There does not seem to be any EventBridge* class in the following link:

https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events

I think a class extending com.amazonaws.services.eventbridge.model.PutEventsRequestEntry may do the work.
Do you have any plan to add something like that?

Thanks!

msailes:
Hi, 

EventBridge is a new service based on CloudWatch Events.

Does the CloudWatch scheduled event object meet your requirements?

https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java

nanodeath:
Unfortunately if you drop in ScheduledEvent instead, you get this:
```
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "version" (class com.amazonaws.services.lambda.runtime.events.ScheduledEvent), not marked as ignorable (8 known properties: "detail", "region", "resources", "detailType", "account", "source", "time", "id"])
```
If you deserialize the object yourself, using Jackson, and add this:
```
.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
it seems to work well enough. Not ideal though.

msailes:
Hi @nanodeath,

I'll add support for EventBridge events after [v4](https://github.com/aws/aws-lambda-java-libs/issues/282) has been released.

Thanks,

Mark

polaskj:
Going to offer a suggestion since this is still open to being implemented. I've seen many uses of `Object` rather than strongly typing fields (as an example `detail` in [ScheduledEvent](https://github.com/aws/aws-lambda-java-libs/blob/e81c824152d3a01f4955b6254b51bde62387eb85/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java#L33)). This leaves the rest of the deserializing efforts up to the caller and can be error-prone.

I'd like to see something more along the lines of this, so a caller-provided generic can represent the detail, instead of just `Object` 

```
[...]

PR #160, opened 2020-09-14
initial class to represent the EventBridge message format.

msailes:
*Issue #, if available:*

https://github.com/aws/aws-lambda-java-libs/issues/109

*Description of changes:*

Adding support for EventBridge / CloudWatch Events

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/EventBridgeEvent.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/EventBridgeEvent.java
@@ -1,0 +1,31 @@
+package com.amazonaws.services.lambda.runtime.events;
+
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.List;
+
+/**
+ * Class to represent the EventBridge event. This is also the CloudWatch Events event format.
+ *
+ * @see <a href="https://docs.aws.amazon.com/lambda/latest/dg/services-cloudwatchevents.html">CloudWatch Events</a>
+ */
+
+@Data
+@Builder(setterPrefix = "with")
+@NoArgsConstructor
+@AllArgsConstructor
+public class EventBridgeEvent {
+
+    private String version;
+    private String id;
+    private String detailType;
+    private String source;
+    private String account;
+    private String time;
+    private String region;
+    private List<String> resources;
+    private Object detail;
+}


aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/EventBridgeEvent.java
```diff
@@ -0,0 +1,31 @@
...
+
+@Data
+@Builder(setterPrefix = "with")
+@NoArgsConstructor
+@AllArgsConstructor
+public class EventBridgeEvent {
+
+    private String version;
+    private String id;
+    private String detailType;
```
aashutoshbane, line 24: Not sure this will get serialized as the field which comes in event looks like detail-type.

msailes: You're absolutely correct. I'm working with the Lambda team to sort this out.

jeromevdl: https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java is the same...

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/EventBridgeEvent.java
```diff
@@ -0,0 +1,31 @@
...
+    private String source;
+    private String account;
+    private String time;
+    private String region;
+    private List<String> resources;
+    private Object detail;
```
jeromevdl, line 30: Why is this an object here while it is a map in https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java.
Actually, I agree with the object more than the map... maybe I miss something.

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/EventBridgeEvent.java
```diff
@@ -0,0 +1,31 @@
...
+ * Class to represent the EventBridge event. This is also the CloudWatch Events event format.
+ *
+ * @see <a href="https://docs.aws.amazon.com/lambda/latest/dg/services-cloudwatchevents.html">CloudWatch Events</a>
+ */
+
+@Data
+@Builder(setterPrefix = "with")
+@NoArgsConstructor
+@AllArgsConstructor
+public class EventBridgeEvent {
```
jeromevdl, line 20: Using the EventBridge "Download code bindings" feature actually provides an AWSEvent (below). Is it useful to have it in the lib, and if so, why not simply use the generic one below?

```
public class AWSEvent<T> {

  @JsonProperty("detail")
  private T detail = null;

  @JsonProperty("detail-type")
  private String detailType = null;

  @JsonProperty("resources")
  private List resources = null;

  @JsonProperty("id")
  private String id = null;

  @JsonProperty("source")
  private String source = null;

  @JsonProperty("time")
  private Date time = null;

  @JsonProperty("region")
  private String region = null;

  @JsonProperty("version")
  private String version = null;

  @JsonProperty("account")
  private String account = null;

// ...
```

msailes: I don't think the Lambda service would be able to handle the typing.

@carlzogh

carlzogh: Unfortunately not at this point, the Java runtime needs to handle `@JsonProperty` annotations internally

msailes:
Closing.

EventBridge is a new service based on CloudWatch Events. The CloudWatch scheduled event object should meet the same requirements.

https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java

aashutoshbane:
> Closing.
> 
> EventBridge is a new service based on CloudWatch Events. The CloudWatch scheduled event object should meet the same requirements.
> 
> https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ScheduledEvent.java

Yeah looks the same but need a workaround for detail-type field :)

Issue #177, opened 2020-10-20
Add S3 Batch Operations Request and Response Models


david-kingsbury-bah:
S3 Batch Operations can invoke a Lambda written in Java. This method of Lambda invocation uses special request and response models. A developer is likely to model the request and response data in Java based on the [batch-ops-invoke-lambda documentation](https://docs.aws.amazon.com/AmazonS3/latest/dev/batch-ops-invoke-lambda.html#batch-ops-invoke-lambda-custom-functions-request).

We could include the Request and Response models in this library. We could even add a little bit of logic to help create the Response model, because most of its data is derived from the Request.

Something like:
```
com.amazonaws.services.lambda.runtime.events.S3BatchOpRequest
com.amazonaws.services.lambda.runtime.events.S3BatchOpResponse
```

msailes:
Hi David, 

I'll look into this.

david-kingsbury-bah:
I was thinking about opening a PR for this issue with some classes that I've already made and tested successfully. I'll see if I can get to that today.

msailes:
That would be great.

This is what I've made, but I haven't had chance to test it yet.

https://github.com/aws/aws-lambda-java-libs/pull/179

Mark

david-kingsbury-bah:
Awesome; testing your branch now.

PR #179, opened 2020-10-21
S3 batch

msailes:
*Issue #, if available:*

https://github.com/aws/aws-lambda-java-libs/issues/177

*Description of changes:*

Event and response objects for S3 Batch operations.

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/S3BatchResponse.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/S3BatchResponse.java
@@ -7,6 +7,13 @@ import lombok.NoArgsConstructor;
 
 import java.util.List;
 
+/**
+ * Event to represent the response which should be returned as part of a S3 Batch custom
+ * action.
+ *
+ * https://docs.aws.amazon.com/AmazonS3/latest/dev/batch-ops-invoke-lambda.html
+ */
+
 @Data
 @Builder(setterPrefix = "with")
 @NoArgsConstructor
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/S3BatchResponse.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/S3BatchResponse.java
@@ -21,7 +21,7 @@ import java.util.List;
 public class S3BatchResponse {
 
     private String invocationSchemaVersion;
-    private String treatMissingKeysAs;
+    private ResultCode treatMissingKeysAs;
     private String invocationId;
     private List<Result> results;
 
@@ -55,5 +55,11 @@ public class S3BatchResponse {
          * 	string. Result strings from failed tasks are ignored.
          */
         PermanentFailure
+    }
+
+    public static S3BatchResponseBuilder fromS3BatchEvent(S3BatchEvent s3BatchEvent) {
+        return S3BatchResponse.builder()
+                .withInvocationId(s3BatchEvent.getInvocationId())
+                .withInvocationSchemaVersion(s3BatchEvent.getInvocationSchemaVersion());
     }
 }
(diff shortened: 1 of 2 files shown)


aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/S3BatchResponse.java
```diff
@@ -0,0 +1,59 @@
...
+@Data
+@Builder(setterPrefix = "with")
+@NoArgsConstructor
+@AllArgsConstructor
+public class S3BatchResponse {
+
+    private String invocationSchemaVersion;
+    private String treatMissingKeysAs;
+    private String invocationId;
+    private List<Result> result;
```
david-kingsbury-bah, line 26:
```suggestion
    private List<Result> results;
```

david-kingsbury-bah: Plural "results"

david-kingsbury-bah (changes requested):
After making the one field name change (result VS results), I was able to successfully run an S3 Batch Operations job that invokes a Lambda written using these models.

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/S3BatchResponse.java
msailes: fixed.

david-kingsbury-bah:
Since we have the ResultCode enum now, I'm thinking S3BatchOpResponse.treatMissingKeysAs should be of type ResultCode.

```
private ResultCode treatMissingKeysAs;
```

### The following are ideas that might go too far for this library; let me know what you think.

We could default treatMissingKeysAs to PermanentFailure, as this is _likely_ the value _most_ people will use.

```
private ResultCode treatMissingKeysAs = ResultCode.PermanentFailure;
```

We could have a special constructor to seed a response instance. These must be set for the response to be considered valid. If one does not want to use this, they can use the other constructors.

```
    public S3BatchResponse(final S3BatchEvent request) {
        this.invocationSchemaVersion = request.getInvocationSchemaVersion();
        this.invocationId = request.getInvocationId();
        this.results = Collections.singletonList(
                Result.builder()
                        .withTaskId(request.getTasks().get(0).getTaskId())
                        .withResultCode(ResultCode.Succeeded)
                        .build()
        );
    }
```

msailes:
Hi David,

Great feedback! What about something like this in the S3BatchResponse.java class.

```java
public static S3BatchResponseBuilder fromS3BatchEvent(S3BatchEvent s3BatchEvent) {
        return S3BatchResponse.builder()
                .withInvocationId(s3BatchEvent.getInvocationId())
                .withInvocationSchemaVersion(s3BatchEvent.getInvocationSchemaVersion());
}
```

I think I'd be nervous adding any default behavior which might have a negative impact. I don't know this area of the platform that well, so I would definitely appreciate any feedback.

Thanks,

Mark

david-kingsbury-bah:
The `fromS3BatchEvent` method looks great!

I think the question now is should `fromS3BatchEvent` go as far as my example and create a single `Result` with the Task ID set.

- S3 Batch Ops expects the response result Task ID to match what was passed in through the request (`S3BatchEvent`). I suspect this allows S3 Batch Ops to do many concurrent invocations and still be able to gather and sort results.
- S3 Batch Ops only sends 1 task, and expects only 1 result. Here is a snippet from their docs
   ```
   # S3 batch operations currently only passes a single task at a time in the array of tasks.
   task = event['tasks'][0]
   ```

I'm on the fence. The API _allows_ for multiple tasks and results, however it is certainly 1 task per event today.

msailes:
In my opinion no. 

If all that was required was the task id it would make sense. But the user will still have to add their own values for resultCode and resultString. So I don't think there is much to be gained.

msailes:
I've also tested this now.

david-kingsbury-bah (approved):

carlzogh (approved):
LGTM!

carlzogh:
Thanks @msailes and @david-kingsbury-bah! :)

david-kingsbury-bah:
Excellent!

Resolved #177

PR #183, opened 2020-10-29
Added conversion to AttributeValue model of original dynamodb sdk

jente-peeraer-cloudway:
*Issue #, if available: 182*

*Description of changes: Added conversion to AttributeValue model of original dynamodb sdk (com.amazonaws.services.dynamodbv2.model.AttributeValue)*

### DIFF
--- a/aws-lambda-java-events-sdk-transformer/pom.xml
+++ b/aws-lambda-java-events-sdk-transformer/pom.xml
@@ -54,10 +54,17 @@
     </dependency>
     <dependency>
       <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-dynamodb</artifactId>
+      <version>1.11.885</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
       <artifactId>aws-lambda-java-events</artifactId>
       <version>3.5.0</version>
       <scope>provided</scope>
     </dependency>
+
 
     <dependency>
       <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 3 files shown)


jente-peeraer-cloudway:
Is there any additional data needed in order to approve this PR?

carlzogh:
Hey @jente-peeraer-cloudway - the initial version of this library was written for SDK v1.

Does version `1.0.0` help with your use-case? https://search.maven.org/artifact/com.amazonaws/aws-lambda-java-events-sdk-transformer/1.0.0

Otherwise we can revisit this and look into supporting both SDK v1 and v2 across the different models in the library, I think if we were to go down that road we'd want to add support for all models not just `AttributeValue`

francis-a:
@carlzogh I'm also running into this issue, it seems like an oversight honestly.

With this SDK we have two versions of `AttributeValue`.

`com.amazonaws.services.lambda.runtime.events.models.dynamodb::AttributeValue` and `com.amazonaws.services.dynamodbv2.model::AttributeValue`. While on a source code level they look identical from the DynamoDB SDK's perspective they aren't. Even more surprising is that the  `com.amazonaws.services.lambda.runtime.events.models.dynamodb` version looks to be a copy and paste from the DynamoSDK since even the documentation references the Dynamo SDK docs.

I looked into your suggestion and it seems like version 1.0.0 of the transformer SDK still only converts between the DyanmoDB SDK V2 models so it doesn't look like ti can help out here. Due to this it seems like the PR from @jente-peeraer-cloudway is still valid. 

I really think this SDK should either use the Dynamo SDK model directly or supply some kind of conversion. Without this for stream converters it's actually easier to go without the typing provided.

carlzogh:
Thanks for highlighting this - I've put in a PR with a different approach to not only support `AttributeValue` but the entire V1 DDB SDK models suite (ref. https://github.com/aws/aws-lambda-java-libs/pull/195)

Will close out this one, thanks again!

PR #193, opened 2020-12-04
Include RIC pr target in GitHub action

stefanbotez:
*Description of changes:*
- Included RIC `pr` target in GitHub action and removed the building commands that were already there since the `pr` target would also build them and would generate conflicts
- Removed unnecessary -it option from codebuild_build.sh because it was causing tty error when running the GitHub action
- Refactored a test that uses a special character to expect failure if the encoding is not set to UTF
- Updated to CentOS:7
- Removed CentOS:6 from the OS compatibility test since it reached EOL

### DIFF
--- a/.github/workflows/maven-build-all.yml
+++ b/.github/workflows/maven-build-all.yml
@@ -31,7 +31,8 @@ jobs:
       run: mvn -B package --file aws-lambda-java-events-sdk-transformer/pom.xml
     - name: Build log4j2 with Maven
       run: mvn -B package --file aws-lambda-java-log4j2/pom.xml
-    - name: Build serialization with Maven
-      run: mvn -B package --file aws-lambda-java-serialization/pom.xml
-    - name: Build runtime-interface-client with Maven
-      run: mvn -B package --file aws-lambda-java-runtime-interface-client/pom.xml
+
+    # Test Runtime Interface Client
+    - name: Run 'pr' target
+      working-directory: ./aws-lambda-java-runtime-interface-client
+      run: make pr
--- a/aws-lambda-java-runtime-interface-client/src/main/jni/Dockerfile.glibc
+++ b/aws-lambda-java-runtime-interface-client/src/main/jni/Dockerfile.glibc
@@ -1,10 +1,11 @@
-# we use centos 6 to build against glibc 2.12
-FROM centos:6
+# we use centos 7 to build against glibc 2.12
+FROM centos:7
 
 # aws-lambda-cpp requires cmake3, it's available in EPEL
 RUN yum install -y epel-release
 RUN yum install -y \
     cmake3 \
+    make \
     gcc \
     gcc-c++ \
     glibc-devel \
--- a/aws-lambda-java-runtime-interface-client/src/main/jni/Dockerfile.glibc
+++ b/aws-lambda-java-runtime-interface-client/src/main/jni/Dockerfile.glibc
@@ -1,4 +1,4 @@
-# we use centos 7 to build against glibc 2.12
+# we use centos 7 to build against glibc 2.17
 FROM centos:7
 
 # aws-lambda-cpp requires cmake3, it's available in EPEL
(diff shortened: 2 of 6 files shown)


aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/CustomerClassLoaderTest.java
```diff
@@ -95,24 +95,35 @@ private Path fakeFileSystem(String[] paths) throws IOException {
...
+                    "/user/path/lib/A.jar",
+                    "/user/path/lib/a.jar",
+                    "/user/path/lib/b.jar",
+                    "/user/path/lib/z.jar",
+                    "/user/path/lib/λ.jar"),
+                    res);
+        } catch(Throwable t) {
+            // this system property is the name of the charset used when encoding/decoding file paths
+            // exception is expected if it is not set to a UTF variant
+            if (!System.getProperty("sun.jnu.encoding").toLowerCase().contains("utf")){
```
carlzogh, line 120: Is this property guaranteed to be set? What happens if it's null?

aws-lambda-java-runtime-interface-client/src/main/jni/Dockerfile.glibc
```diff
@@ -1,10 +1,11 @@
-# we use centos 6 to build against glibc 2.12
-FROM centos:6
+# we use centos 7 to build against glibc 2.12
```
carlzogh, line 1: With CentOS 7 as a base, we are now building against `glibc` 2.17 - as a result, this comment and requirements in `README.md` should be updated to reflect this too.

Building against `glibc` 2.17 effectively means that the minimum supported version is 2.17 now

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/CustomerClassLoaderTest.java
stefanbotez: Correct. There is the possibility to be null. Treating the case in next commit.

carlzogh (approved):

edhzsz (approved):
LGTM!

.github/workflows/maven-build-all.yml
```diff
@@ -31,7 +31,8 @@ jobs:
...
       run: mvn -B package --file aws-lambda-java-log4j2/pom.xml
-    - name: Build serialization with Maven
-      run: mvn -B package --file aws-lambda-java-serialization/pom.xml
-    - name: Build runtime-interface-client with Maven
-      run: mvn -B package --file aws-lambda-java-runtime-interface-client/pom.xml
+
+    # Test Runtime Interface Client
+    - name: Run 'pr' target
+      working-directory: ./aws-lambda-java-runtime-interface-client
+      run: make pr
```
carlzogh, line 38: Looks like this actually takes a while - I suggest we create a separate workflow that would only trigger if changes were made for the RIC or any of its dependent libraries (`aws-lambda-java-core` and `aws-lambda-java-serialization`).

ref. https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-including-paths

carlzogh (changes requested):

.github/workflows/maven-build-all.yml
carlzogh: We could even take the opportunity to define a workflow per package as there is probably no need to build them all on every change.

stefanbotez: Creating a follow up PR for restructuring the workflow.

carlzogh: Sounds good!

carlzogh (approved):

PR #194, opened 2020-12-07
Stop building gcc from source

carlzogh:
*Description of changes:*
This PR:
- Changes the RIC's build process to use gcc from `yum` as opposed to building it from source (as it's now a modern supported version after updating to CentOS 7 done [here](https://github.com/aws/aws-lambda-java-libs/pull/193))
- Removes unnecessary `user.dir` overrides in `CustomerClassLoaderTest`
- Adds glibc OS test in test-smoke (now also testing `amazoncorretto` version `11` - previously we were only testing Alpine with musl)

### DIFF
--- a/aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/CustomerClassLoaderTest.java
+++ b/aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/CustomerClassLoaderTest.java
@@ -2,9 +2,7 @@
 
 package com.amazonaws.services.lambda.runtime.api.client;
 
-import org.junit.jupiter.api.AfterEach;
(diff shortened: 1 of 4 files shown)


stefanbotez (approved):
LGTM!

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/CustomerClassLoaderTest.java
```diff
@@ -20,18 +18,6 @@
 import static org.junit.jupiter.api.condition.OS.MAC;
 
 public class CustomerClassLoaderTest {
-    private static final String USER_WORKING_DIR = "user/path";
```
edhzsz, line 23: Feels like this changes should had been done in a different PR.

edhzsz (approved):

PR #196, opened 2020-12-08
Separate GitHub actions workflows

stefanbotez:
*Description of changes:*
- GitHub actions workflow currently builds all packages on push and pull request no matter of the package changed
- Separated the GitHub actions workflows to run based on the package changed. It builds the package changed and the packages that depend on that package. There could be some duplication builds if multiple packages are changed in a single pull request, but for most cases it should not be a problem since before including the RIC, building all the other packages took around 1 minute.

### DIFF
--- a/.github/workflows/aws-lambda-java-events-sdk-transformer.yml
+++ b/.github/workflows/aws-lambda-java-events-sdk-transformer.yml
@@ -1,0 +1,31 @@
+# This workflow will be triggered if there will be changes to 
+# aws-lambda-java-events-sdk-transformer package and it builds the package.
+
+name: Java CI aws-lambda-java-events-sdk-transformer
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+    - 'aws-lambda-java-events-sdk-transformer/**'
+  pull_request:
+    branches: [ '*' ]
+    paths:
+    - 'aws-lambda-java-events-sdk-transformer/**'
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    
+    # Install base module
+    - name: Build events-sdk-transformer with Maven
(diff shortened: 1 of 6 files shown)


.github/workflows/aws-lambda-java-events-sdk-transformer.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    
+    # Install base module
+    - name: Build events-sdk-transformer with Maven
+      run: mvn -B install --file aws-lambda-java-events-sdk-transformer/pom.xml
```
carlzogh, line 30: I think we should also be installing dependency modules here and in other workflows (ie. `aws-lambda-java-events` in this case).

If a PR contains changes for both these libraries and includes a version bump to the base package (`aws-lambda-java-events`) and a dependent change in the consumer library (`aws-lambda-java-events-sdk-transformer`), the workflow would fail as it wouldn't be able to resolve and install the new version from remote repositories.

carlzogh (changes requested):

.github/workflows/aws-lambda-java-serialization.yml
```diff
@@ -26,6 +26,9 @@ jobs:
         java-version: 1.8
     
     # Install base module
+    - name: Build events with Maven
+      run: mvn -B install --file aws-lambda-java-events/pom.xml
+    # Install target module
     - name: Build serialization with Maven
       run: mvn -B install --file aws-lambda-java-serialization/pom.xml
```
carlzogh: nit: Why install this target module when others are packaged instead?

stefanbotez: I did this because the next step is to test `aws-lambda-java-runtime-interface-client`, which is a consumer of this package and I thought it would be good for `aws-lambda-java-serialization` to  be installed in the local repository.

carlzogh: Right, gotcha 👍

carlzogh (approved):

PR #200, opened 2020-12-10
Revise GitHub workflows

stefanbotez:
*Description of changes:*
- The `pr` target of the Runtime Interface Client already has the commands to install its dependencies `aws-lambda-java-core` and `aws-lambda-java-serialization`. Removed them from the GitHub workflows to avoid duplication and most importantly to avoid errors like`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:3.2.1:shade (default) on project aws-lambda-java-serialization: Error creating shaded jar: duplicate entry` when trying to install twice `aws-lambda-java-serialization`.


.github/workflows/aws-lambda-java-serialization.yml
```diff
@@ -28,9 +28,9 @@ jobs:
     # Install base module
     - name: Build events with Maven
       run: mvn -B install --file aws-lambda-java-events/pom.xml
-    # Install target module
+    # Package target module
     - name: Build serialization with Maven
```
carlzogh, line 32: nit: Looking at this I think we should rename the workflow steps in a way that more accurately depicts what the workflow is doing with the module - eg. `s/Build/Package` and `s/Build/Install` when appropriate.

carlzogh (approved):

carlzogh (approved):

PR #237, opened 2021-04-09
use environment variable _HANDLER if no args

sdelamo:
Right now `AWSLambda` main method expects the handler to be supplied as an argument. This PR uses the environment variable `_HANDLER` if no argument is provided. 

see: https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

> `_HANDLER` – The handler location configured on the function.

### DIFF
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambda.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambda.java
@@ -185,7 +185,7 @@ public class AWSLambda {
 
     public static void main(String[] args) {
         // TODO validate arguments, show usage
-        startRuntime(args[0]);
+        startRuntime(args.length > 0 ? args[0] : getEnvOrExit("_HANDLER"));
     }
 
     private static void startRuntime(String handler) {


msailes:
Hi Sergio,

Does this enable further work you are doing in Micronaut?

Thanks,

Mark

sdelamo:
Yes. it would simplify Micronaut integration. I have been checking using the `aws-lambda-java-runtime-interface-client` for deploying a GraalVM native image of a  Micronaut AWS lambda application to a AWS Lambda custom runtime. 

Do you see any downside? As far as I understand, `_HANDLER` env. variable will be present.

carlzogh:
Hey @sdelamo - thanks for your contribution.

For Runtime Interface Client libraries, we have opted out of exposing the concept of the `_HANDLER` environment variable as it is currently only a concept that applies to ZIP (non-container image) functions. Container images in Lambda are characterised by the combination of the `ENTRYPOINT` and `CMD` directives that signal to Lambda what it should run within the container, whereas ZIP functions instead use a Handler configuration to achieve this.

I would see the `_HANDLER` variable definition as being part of the bootstrap around the Runtime Interface Client for non-image functions that would be responsible for passing the variable's value to the RIC as a command-line argument. Are there limitations in the Micronaut integration where this cannot happen? Is there any guidance we could provide to help achieve this differently?

sdelamo:
I think we can work around `_HANDLER` environment variable not being there. However, we would probably need to generate a class which calls your class with the handler class as the first argument or change the way we invoke the GraalVM native image supplying the Handler as an argument. But, I think we could work around it. It would be less elegant but I think we can solve it. 

However, I still find confusing that you don't expose the variable. This documentation https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html should clarify for which environments are the variables present and for which there are not. Moreover, I don't see the harm in exposing env variable as documented. The user would have the ability to define the  Handler in the AWS Console and for some custom runtime it maybe useful information that we can provide.

msailes:
Closing after discussions with @sdelamo and @carlzogh

Issue #207, opened 2020-12-23
Unrecognized field "version" in class com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent


Alappin:
A deployed lambda written in java fails in API Gateway due to a missing version field. Error message is shown below:

`Unrecognized field "version" (class com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent), not marked as ignorable (12 known properties: "requestContext", "body", "queryStringParameters", "path", "headers", "httpMethod", "multiValueQueryStringParameters", "resource", "pathParameters", "isBase64Encoded", "stageVariables", "multiValueHeaders"])`

It seems the latest java lambda events is missing the "version" field hence the event cannot be serialized without throwing an exception.

msailes:
Hi Alappin,

Are you using version 2 of the API Gateway proxy integration?

If you are that uses a different event which does have `version`

https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2HTTPEvent.java

Mark

Alappin:
Hi @msailes 

 For my AGW HTTP API, in my code, I am using the (APIGatewayProxyRequestEvent) with the format: 

`
{
    "resource": "Resource path",
    "path": "Path parameter",
    "httpMethod": "Incoming request's method name",
    ...
}
`

When I test this locally, I send requests with the below form to the lambda container image and this works

`{
	"path":"/api",
	"httpMethod": "POST",
}
`

However, when I deploy the lambda and have my Version 1 integration set in AGW, I get requests of the form:

`{
        "version": "1.0",
	"path":"/api",
	"httpMethod": "POST",
}
`

AGW adds the "version" tag regardless of whether its configured for V1 or V2 integration causing the lambda to fail because the "APIGatewayProxyRequestEvent" cannot serialize a "version" property.

So I think, either AGW should be fixed to exclude "version" if configured for V1 or the lambda events package should add the "version" property so that it does not break when deployed. I hope my problem is clear.

Alappin:
Also checking the documentation on https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format

[...]

PR #258, opened 2021-07-09
Missing version field

msailes:
*Issue #, if available:*
https://github.com/aws/aws-lambda-java-libs/issues/207

*Description of changes:*
This is a little complex due to the various different modes of operation around REST and HTTP APIs, with the old (v1) and new (v2) message formats.

- REST APIs should use `APIGatewayProxyRequestEvent` a version won't be set. This doesn't cause a problem, but `getVersion()` does return a null.
- HTTP APIs using v1 and the `APIGatewayProxyRequestEvent` will have "1.0" set for the version.
- HTTP APIs using v2 and the `APIGatewayProxyRequestEvent` will have "2.0" set for the version.
- HTTP APIs using v2 should use the `APIGatewayV2HTTPEvent` to take advantage of the changes.

### DIFF
--- a/.github/workflows/aws-lambda-java-serialization.yml
+++ b/.github/workflows/aws-lambda-java-serialization.yml
@@ -30,4 +30,4 @@ jobs:
       run: mvn -B install --file aws-lambda-java-events/pom.xml
     # Package target module
     - name: Package serialization with Maven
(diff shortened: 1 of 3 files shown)


msailes:
I've tested all these variations on AWS.

carlzogh (approved):
LGTM

PR #260, opened 2021-07-22
Add headers in KafkaEventRecord

ianpyw:
*Issue #, if available:*

*Description of changes:* This change add headers in KafkaEventRecord 

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/KafkaEvent.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/KafkaEvent.java
@@ -42,6 +42,7 @@ public class KafkaEvent {
         private String timestampType;
         private String key;
         private String value;
+        private List<Map<String, byte[]>> headers;
     }
 
     @Data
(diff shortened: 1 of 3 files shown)


zsombor-balogh (approved):
Lgtm

smirnoal (approved):
lgtm

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/KafkaEvent.java
```diff
@@ -42,6 +42,7 @@
         private String timestampType;
         private String key;
         private String value;
+        private List<Map<String, byte[]>> headers;
```
msailes, line 45: Would this be better as a `String` rather than a `byte[]`?

ianpyw: The byte[] is [defined by Kafka](https://kafka.apache.org/26/javadoc/index.html?org/apache/kafka/common/header/Headers.html), I think we should pass what we got from Kafka to customer.

Issue #168, opened 2020-10-07
`record.getUserIdentity()`


guilhermecvm:
Hi, trying to convert `DynamodbEvent` to `List<Record>`, however when the lambda gets excuted, seems like `record.getUserIdentity()` returns `null`, therefore breaking `toIdentityV2`. Is it the case that the user identity is nullable and a safety check is needed?

```kotlin
fun handleRequest(event: DynamodbEvent, context: Context) {
  DynamodbEventTransformer.toRecordsV2(event)
}
```

```
java.lang.NullPointerException
at com.amazonaws.services.lambda.runtime.events.transformers.dynamodb.DynamodbIdentityTransformer.toIdentityV2(DynamodbIdentityTransformer.java:9)
at com.amazonaws.services.lambda.runtime.events.transformers.dynamodb.DynamodbRecordTransformer.toRecordV2(DynamodbRecordTransformer.java:19)
at java.base/java.util.stream.ReferencePipeline$3$1.accept(Unknown Source)
at java.base/java.util.stream.ReferencePipeline$2$1.accept(Unknown Source)
at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(Unknown Source)
at java.base/java.util.stream.AbstractPipeline.copyInto(Unknown Source)
at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source)
at java.base/java.util.stream.AbstractPipeline.evaluate(Unknown Source)
at java.base/java.util.stream.ReferencePipeline.collect(Unknown Source)
at com.amazonaws.services.lambda.runtime.events.transformers.DynamodbEventTransformer.toRecordsV2(DynamodbEventTransformer.java:19)
```

carlzogh:
Hey @guilhermecvm, thanks for reporting this!

I've put in a fix for this issue [here](https://github.com/aws/aws-lambda-java-libs/pull/169)

carlzogh:
The fix for this was released as part of `aws-lambda-java-events-sdk-transformer` version `2.0.6` and is now available through Maven https://search.maven.org/artifact/com.amazonaws/aws-lambda-java-events-sdk-transformer/2.0.6/jar

Thanks again!

guilhermecvm:
Awesome, that was quick. Thanks @carlzogh

PR #264, opened 2021-07-31
Fix NPE when user identity is null for SDK v1

nikrush13:
*Issue #, if available:*
#168 same issue as this, which was fixed for SDK v2 in #169 

*Description of changes:*
Fixes a potential NPE when `userIdentity` is null and added a null check to the OldImage and NewImage for dynamo stream records.


carlzogh (approved):
LGTM - thanks!

PR #270, opened 2021-09-15
Avoid ByteArrayOutputStream copy

richarddd:
*Issue #, if available:*

*Description of changes:*
Introduces an extention class of ByteArrayOutputStream which contains a method to return the raw underlying byte array. This way we can avoid a `System.arraycopy` for the entire stream to save memory and (possibly) performance.

Another possible improvement could be to use a [ByteBuffer backed OutputStream](https://gist.github.com/ncruces/5a1970c4fcfe5c91998046195a6dc70f) or [FastByteArrayOutputStream](https://github.com/spring-projects/spring-framework/blob/main/spring-core/src/main/java/org/springframework/util/FastByteArrayOutputStream.java) which is better at expansion.

Also added casts to Buffer since some methods of ByteBuffer does not exists in older JDKs casing tests to fail in some cases.
Affected methods are position(), flip(), clear() etc
See https://issues.apache.org/jira/browse/MRESOLVER-85 or
https://stackoverflow.com/questions/61267495/exception-in-thread-main-java-lang-nosuchmethoderror-java-nio-bytebuffer-flip 


msailes:
What are the performance improvements of implementing this?

richarddd:
> What are the performance improvements of implementing this?

This really depends on the payload size. For most if the cases it will be negligible since the payload is very small and Java is using a native `System.arraycopy` call. But there are cases when lambda returns binary payloads where we can save a few milliseconds.

msailes:
I think that if this is positioned as a optimisation then it would be good to get some data for various workload patterns.

msailes:
@smirnoal can you have a look at this?

carlzogh (approved):

smirnoal:
I'd avoid merging it. The optimisation only covers exceptional cases, where the payload size is not expected to be that huge. I even think that System.arrayCopy is not so bad in terms of performance for regular 6Mb payload. Would be nice to have supporting data to justify the increase in code complexity

PR #275, opened 2021-09-29
Add ARM64 architecture support

keshayad:
Description of changes:

Add support for ARM64 architecture.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


carlzogh (approved):

edhzsz (approved):
LGTM!

richarddd:
This does not work since the jar does not contain prebuilt libraries for both x64 and arm.
Currently: https://github.com/aws/aws-lambda-java-libs/blob/d1b016a6e8b90526e60d46b266031c218318541f/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java#L16-L19 looks for only two libraries to load when https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-runtime-interface-client/src/main/jni/build-jni-lib.sh only builds for musl and gclib but not for arm

Futhermore, it would be better to build different flavours of `aws-lambda-java-runtime-interface-client` to accomodate the diffrent architectures and standard libraries since this code makes assumptions and relies on exception handling to load the correct library: https://github.com/aws/aws-lambda-java-libs/blob/d1b016a6e8b90526e60d46b266031c218318541f/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java#L23-L39

It would be good to have:

- aws-lambda-java-runtime-interface-client-musl-x64
- aws-lambda-java-runtime-interface-client-gclib-x64
- aws-lambda-java-runtime-interface-client-musl-arm
- aws-lambda-java-runtime-interface-client-gclib-arm

Rather than one jar containing all libraries

PR #234, opened 2021-03-27
fix issue #220, using treemap

jeromevdl:
*Issue #, if available:* #220

*Description of changes:* Using TreeMap (and CASE_INSENSITIVE_ORDER initialisation) within a custom Header object.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #220, opened 2021-02-03
API Gateway Proxy: headers are case sensitive


jordanms:
By standard headers is case insensitive, but `APIGatewayProxyRequestEvent` is instantiated using a regular Map (probably a Hashmap) which causes headers to be case sensitive. I couldn't find the code where it is instantiated, so Idk exactly what is going on there.

This is a similar problem, but for Go language: https://github.com/aws/aws-lambda-go/issues/117

Issue #282, opened 2021-12-08
v4 Events Lib - Major Version Change


msailes:
The next major version of the `aws-lambda-java-events` library is now being worked on. As part of this we are considering changes around serialization. Our goals are to make it easier for customers to handle different serialization scenarios, reduce package size and to more easily add new events to the library.

Our current plans include enabling customers to bring their own serializer by implementing the [PojoSerializer](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/runtime/serialization/PojoSerializer.java) interface and including a provider-configuration file under `META-INF/services`.

To make it easier for new event objects to be added we are exploring ways to remove the use of mixins in the serialization library.

For example the [SecretsManagerRotationEvent](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/SecretsManagerRotationEvent.java) uses the [SecretsManagerRotationEventMixin.java](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/runtime/serialization/events/mixins/SecretsManagerRotationEventMixin.java) to change upper case attributes `Step`, `SecretId` and `ClientRequestToken` to lower case ones. This is required because Jackson expects lower case attribute names. Instead Jackson annotations will be added directly to the event fields which require them.

At this time we’ll also change the existing event objects to all follow the same design. Lombok annotations `@Data`, `@NoArgsConstructor`, `@Builder(setterPrefix = "with")`, `@AllArgsConstructor` and `@Singular`(when needed) will be added to events without them.

The `joda-time` dependency will be removed in this release.

msailes:
@ilopmar, @olegz, @GoodforGod, @sdelamo, @FieryCod

drissamri:
I'm not entirely aware on how the internals work right now. In my experiments around cold start performance I have always opted to use [Jackson-jr](https://github.com/FasterXML/jackson-jr) instead of regular Jackson since it positively impacts the cold start. 

Will this change make it possible to have internal serialization also leverage faster frameworks and be replace the current (slower) Jackson lib?

sdelamo:
For a sample SQS event such as: 

```json
{
  "Records": [
    {
      "messageId": "19dd0b57-b21e-4ac1-bd88-01bbb068cb78",
[...]

PR #298, opened 2021-12-22
#220 Fix case-sensitivity of HTTP headers in models

parawanderer:
*Issue #, if available:* #220

*Description of changes:*

This solution builds upon @jeromevdl 's solution of PR [234](https://github.com/aws/aws-lambda-java-libs/pull/234) to fix the
issue of headers being case sensitive. Fixes @carlzogh 's comment
regarding breaking changes in setHeaders.

This adds some extra unit tests, and reorders some of the added unit test assertions of the original PR, too


msailes:
Do you think that `HttpHeadersUtil.mergeOrReplace()` should be a function within the `HttpHeaders` class, instead of within it's own class.

Thoughts @carlzogh @andclt @jeromevdl

parawanderer:
> Do you think that `HttpHeadersUtil.mergeOrReplace()` should be a function within the `HttpHeaders` class, instead of within it's own class.
> 
> Thoughts @carlzogh @andclt @jeromevdl

Personally I could agree to that too, since you can just as easily have it be package-private on a method level that way and that was the reason I thought of moving it to a package-private class to begin with

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ApplicationLoadBalancerRequestEvent.java
```diff
@@ -40,8 +41,8 @@
     private String path;
     private Map<String, String> queryStringParameters;
     private Map<String, List<String>> multiValueQueryStringParameters;
-    private Map<String, String> headers;
-    private Map<String, List<String>> multiValueHeaders;
+    private HttpHeaders<String> headers;
+    private HttpHeaders<List<String>> multiValueHeaders;
```
carlzogh, line 45: I believe this is a breaking change for some consumers - I suggest we include it as part of the next major version of this library (v4). See https://github.com/aws/aws-lambda-java-libs/issues/282 where changes are being discussed

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2WebSocketResponse.java
```diff
@@ -12,8 +14,8 @@
 
   private boolean isBase64Encoded = false;
   private int statusCode;
-  private Map<String, String> headers;
-  private Map<String, String[]> multiValueHeaders;
+  private HttpHeaders<String> headers;
+  private HttpHeaders<String[]> multiValueHeaders;
```
carlzogh, line 18: If we're doing this change as part of v4, we should also change `String[]` to `List<String>` for consistency with other data models

aws-lambda-java-tests/src/test/java/com/amazonaws/services/lambda/runtime/tests/EventLoaderTest.java
```diff
@@ -26,7 +26,6 @@ public void testLoadApiGatewayRestEvent() {
 
         assertThat(event).isNotNull();
         assertThat(event.getBody()).isEqualTo("Hello from Lambda!");
-        assertThat(event.getHeaders()).hasSize(2);
```
carlzogh, line 29: Why remove this? Can we not include an equivalent assertion for `HttpHeaders`?

msailes:
Hi @parawanderer,

Can this be updated at written as a PR against the [V4](https://github.com/aws/aws-lambda-java-libs/tree/events-v4-serialization-v2) branch?

aws-lambda-java-tests/src/test/java/com/amazonaws/services/lambda/runtime/tests/EventLoaderTest.java
parawanderer: I think I removed this because I created a new test class dedicated solely to testing headers in `aws-lambda-java-tests/src/test/java/com/amazonaws/services/lambda/runtime/tests/HeadersTest.java`, and the header check here did not seem relevant to the name of this set of tests `EventLoaderTest`. At least that's the reasoning that would make sense to me right now.

But since later classes in here do the header test too, right now I'd think there's no real benefit to removing it either. So I'll add it back!

parawanderer:
Hi @msailes

I've taken a look and I believe I merged it in with the "events-v4-serialization-v2" branch, changed things in line with previous suggestions in this thread and updated all relevant tests/usages. 
Please feel free to check this again for any more issues.

PR #300, opened 2021-12-30
add support for ARM based platforms

muellerc:
*Description of changes:
The current version 2.0.0 doesn't support ARM based platforms, as the native code is compiled for x86 only.
This change extends the build process to [cross-] compile the native code for x86 and ARM. When executed on the target platform, it detects the underlying architecture (x86 or ARM) and loads these native libraries only.

### DIFF
--- a/aws-lambda-java-events/pom.xml
+++ b/aws-lambda-java-events/pom.xml
@@ -34,6 +34,7 @@
   <properties>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
+    <lombok.version>1.18.22</lombok.version>
   </properties>
 
   <distributionManagement>
@@ -72,7 +73,7 @@
     <dependency>
       <groupId>org.projectlombok</groupId>
       <artifactId>lombok</artifactId>
-      <version>1.18.16</version>
+      <version>${lombok.version}</version>
       <scope>provided</scope>
     </dependency>
   </dependencies>
@@ -108,7 +109,7 @@
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-source-plugin</artifactId>
-            <version>2.2.1</version>
+            <version>3.2.1</version>
             <executions>
               <execution>
                 <id>attach-sources</id>
--- a/aws-lambda-java-events/pom.xml
+++ b/aws-lambda-java-events/pom.xml
@@ -159,6 +159,29 @@
               <autoReleaseAfterClose>false</autoReleaseAfterClose>
             </configuration>
           </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-resources-plugin</artifactId>
+            <version>3.2.0</version>
+            <configuration>
+              <encoding>UTF-8</encoding>
+            </configuration>
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <version>3.8.1</version>
+            <configuration>
+              <annotationProcessorPaths>
+                <path>
+                  <groupId>org.projectlombok</groupId>
+                  <artifactId>lombok</artifactId>
+                  <version>${lombok.version}</version>
+                </path>
+              </annotationProcessorPaths>
+              <encoding>UTF-8</encoding>
+            </configuration>
+          </plugin>
         </plugins>
       </build>
     </profile>
--- a/aws-lambda-java-events/pom.xml
+++ b/aws-lambda-java-events/pom.xml
@@ -5,7 +5,7 @@
 
   <groupId>com.amazonaws</groupId>
   <artifactId>aws-lambda-java-events</artifactId>
-  <version>3.10.0</version>
+  <version>3.11.0</version>
   <packaging>jar</packaging>
 
   <name>AWS Lambda Java Events Library</name>
--- a/aws-lambda-java-runtime-interface-client/README.md
+++ b/aws-lambda-java-runtime-interface-client/README.md
@@ -13,7 +13,7 @@ You can include this package in your preferred base image to make that base imag
 
 Choose a preferred base image. The Runtime Interface Client is tested on Amazon Linux, Alpine, Ubuntu, Debian, and CentOS. The requirements are that the image is:
 
-* built for x86_64 and ARM
+* built for x86_64 and ARM64
 * contains Java >= 8
 * contains glibc >= 2.17 or musl
 
--- a/aws-lambda-java-runtime-interface-client/README.md
+++ b/aws-lambda-java-runtime-interface-client/README.md
@@ -13,7 +13,7 @@ You can include this package in your preferred base image to make that base imag
 
 Choose a preferred base image. The Runtime Interface Client is tested on Amazon Linux, Alpine, Ubuntu, Debian, and CentOS. The requirements are that the image is:
 
-* built for x86_64
+* built for x86_64 and ARM
 * contains Java >= 8
 * contains glibc >= 2.17 or musl
 
@@ -70,7 +70,7 @@ pom.xml
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-lambda-java-runtime-interface-client</artifactId>
-      <version>2.0.0</version>
+      <version>2.1.0</version>
     </dependency>
   </dependencies>
   <build>
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
@@ -13,15 +13,17 @@ import java.nio.file.StandardCopyOption;
  */
 class NativeClient {
     private static final String nativeLibPath = "/tmp/.aws-lambda-runtime-interface-client";
+    private static final String architecturePathSuffix = "/" + getArchIdentifier();
     private static final String[] libsToTry = {
-            "/aws-lambda-runtime-interface-client.glibc.so",
-            "/aws-lambda-runtime-interface-client.musl.so",
+            "aws-lambda-runtime-interface-client.glibc.so",
+            "aws-lambda-runtime-interface-client.musl.so",
     };
     private static final Throwable[] exceptions = new Throwable[libsToTry.length];
     static {
             boolean loaded = false;
             for (int i = 0; !loaded && i < libsToTry.length; ++i) {
-                try (InputStream lib = NativeClient.class.getResourceAsStream(libsToTry[i])) {
+                try (InputStream lib = NativeClient.class.getResourceAsStream(
+                        Paths.get(architecturePathSuffix, libsToTry[i]).toString())) {
                     Files.copy(lib, Paths.get(nativeLibPath), StandardCopyOption.REPLACE_EXISTING);
                     System.load(nativeLibPath);
                     loaded = true;
@@ -44,10 +46,25 @@ class NativeClient {
             initializeClient(userAgent.getBytes());
     }
 
+    /**
+     * @return a string describing the detected architecture the RIC is executing on
+     * @throws RuntimeException
+     */
+    static String getArchIdentifier() {
+        String arch = System.getProperty("os.arch");
+
+        if (arch.matches("^(x8664|amd64|ia32e|em64t|x64|x86_64|x8632|x86|i[3-6]86|ia32|x32)$")) {
+            return "x86";
+        } else if (arch.matches("^(armeabi.*|arm64.*|aarch64.*|arm)$")) {
+            return "arm";
+        }
+
+        throw new RuntimeException("architecture not supported: " + arch);
+    }
+
     static native void initializeClient(byte[] userAgent);
 
     static native InvocationRequest next();
 
     static native void postInvocationResponse(byte[] requestId, byte[] response);
-
 }
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
@@ -50,15 +50,15 @@ class NativeClient {
 
     /**
      * @return a string describing the detected architecture the RIC is executing on
-     * @throws RuntimeException
+     * @throws UnknownPlatformException
      */
     static String getArchIdentifier() {
         String arch = System.getProperty("os.arch");
 
         if (arch.matches(supported_x86_architectures)) {
-            return "x86";
+            return "x86_64";
         } else if (arch.matches(supported_arm_architectures)) {
-            return "arm";
+            return "arm64";
         }
 
         throw new UnknownPlatformException("architecture not supported: " + arch);
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
@@ -14,6 +14,8 @@ import java.nio.file.StandardCopyOption;
 class NativeClient {
     private static final String nativeLibPath = "/tmp/.aws-lambda-runtime-interface-client";
     private static final String architecturePathSuffix = "/" + getArchIdentifier();
+    private static final String supported_arm_architectures = "^(aarch64)$";
+    private static final String supported_x86_architectures = "^(amd64)$";
     private static final String[] libsToTry = {
             "aws-lambda-runtime-interface-client.glibc.so",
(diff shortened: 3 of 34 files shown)


muellerc:
This change can be tested using this sample project: https://github.com/muellerc/lambda-java-ric-platform-test

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
```diff
@@ -44,10 +46,25 @@
...
+    static String getArchIdentifier() {
+        String arch = System.getProperty("os.arch");
+
+        if (arch.matches("^(x8664|amd64|ia32e|em64t|x64|x86_64|x8632|x86|i[3-6]86|ia32|x32)$")) {
+            return "x86";
+        } else if (arch.matches("^(armeabi.*|arm64.*|aarch64.*|arm)$")) {
+            return "arm";
+        }
+
+        throw new RuntimeException("architecture not supported: " + arch);
```
smirnoal, line 62: let us [not throw](https://wiki.sei.cmu.edu/confluence/display/java/ERR07-J.+Do+not+throw+RuntimeException%2C+Exception%2C+or+Throwable) RuntimeException

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
```diff
@@ -44,10 +46,25 @@
...
+     * @throws RuntimeException
+     */
+    static String getArchIdentifier() {
+        String arch = System.getProperty("os.arch");
+
+        if (arch.matches("^(x8664|amd64|ia32e|em64t|x64|x86_64|x8632|x86|i[3-6]86|ia32|x32)$")) {
+            return "x86";
+        } else if (arch.matches("^(armeabi.*|arm64.*|aarch64.*|arm)$")) {
+            return "arm";
+        }
```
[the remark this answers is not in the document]
msailes: This is how CRT has implemented it 

https://github.com/awslabs/aws-crt-java/blob/main/src/main/java/software/amazon/awssdk/crt/CRT.java#L83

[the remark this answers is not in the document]
msailes: Does the work have any environment variables which could be used?

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
carlzogh: Suggest wrapping in new exception eg. `UnknownPlatformException`

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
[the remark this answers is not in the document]
carlzogh: We should further restrict this to `x86_64` and `arm64` exclusively, using a similar distinction as done in the CRT. We don't want to consider 32-bit architectures as supported under `x86` as they're not really.

aws-lambda-java-runtime-interface-client/src/main/jni/build-jni-lib.sh
```diff
@@ -7,10 +7,18 @@ SRC_DIR=$(dirname "$0")
...
+docker build --platform=linux/amd64 -f "${SRC_DIR}/Dockerfile.musl" --build-arg CURL_VERSION=${CURL_VERSION} -t lambda-java-jni-lib-musl-x86 "${SRC_DIR}"
+docker run --rm --entrypoint /bin/cat lambda-java-jni-lib-musl-x86 /src/aws-lambda-runtime-interface-client.so > "${DST_DIR}"/classes/x86/aws-lambda-runtime-interface-client.musl.so
 
+# compile the native library for ARM
+mkdir "${DST_DIR}"/classes/arm
+docker build --platform=linux/arm64/v8 -f "${SRC_DIR}/Dockerfile.glibc" --build-arg CURL_VERSION=${CURL_VERSION} -t lambda-java-jni-lib-glibc-arm "${SRC_DIR}"
+docker run --rm --entrypoint /bin/cat lambda-java-jni-lib-glibc-arm /src/aws-lambda-runtime-interface-client.so > "${DST_DIR}"/classes/arm/aws-lambda-runtime-interface-client.glibc.so
+
+docker build --platform=linux/arm64/v8 -f "${SRC_DIR}/Dockerfile.musl" --build-arg CURL_VERSION=${CURL_VERSION} -t lambda-java-jni-lib-musl-arm "${SRC_DIR}"
+docker run --rm --entrypoint /bin/cat lambda-java-jni-lib-musl-arm /src/aws-lambda-runtime-interface-client.so > "${DST_DIR}"/classes/arm/aws-lambda-runtime-interface-client.musl.so
```
carlzogh, line 24: Room for refactoring and code reuse

carlzogh (changes requested):
Thanks again for the contribution - I've made a few changes on top of those in  I've pushed a commit that does this: https://github.com/carlzogh/aws-lambda-java-libs/commit/0b16ecde364fe03ea5036a9301581a1cfb9b9797

@muellerc what do you think about including these changes in your PR? (I tried pushing to your fork but it seems I don't have permissions to do so)

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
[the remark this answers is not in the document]
msailes: 👍🏻

aws-lambda-java-runtime-interface-client/src/main/jni/build-jni-lib.sh
muellerc: Sure @carlzogh . Let me refactor the code as you proposed and come back with an updated pull request later this week, maybe tomorrow.

muellerc:
@carlzogh , I already managed to incorporate the proposed improvements. Let me know if there is something else I can do.

msailes:
I tried looking at why this build was failing, but I can't work it out.

I can only think it's something to do with the phase in which lombok code is generated.

https://github.com/aws/aws-lambda-java-libs/blob/bd490cb4a5a415fcb9df67f95a854b37dd38bd9a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/S3BatchResponse.java#L60

I can't see what else it would be.

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
```diff
@@ -44,10 +48,25 @@
...
+
+        if (arch.matches(supported_x86_architectures)) {
+            return "x86";
+        } else if (arch.matches(supported_arm_architectures)) {
+            return "arm";
+        }
+
+        throw new UnknownPlatformException("architecture not supported: " + arch);
+    }
+
```
carlzogh, line 66: Could you please use the naming/filter patterns suggested in https://github.com/carlzogh/aws-lambda-java-libs/commit/0b16ecde364fe03ea5036a9301581a1cfb9b9797? I'd also rather we don't use `x86` or `arm` but `x86_64` and `arm64` all around.

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
```diff
@@ -44,10 +48,25 @@
             initializeClient(userAgent.getBytes());
     }
 
+    /**
+     * @return a string describing the detected architecture the RIC is executing on
+     * @throws RuntimeException
```
carlzogh, line 53: `RuntimeException` -> `UnknownPlatformException`

aws-lambda-java-runtime-interface-client/src/main/jni/build-jni-lib.sh
```diff
@@ -7,10 +7,14 @@ SRC_DIR=$(dirname "$0")
...
 
-docker build -f "${SRC_DIR}/Dockerfile.musl" --build-arg CURL_VERSION=${CURL_VERSION} -t lambda-java-jni-lib-musl "${SRC_DIR}"
-docker run --rm --entrypoint /bin/cat lambda-java-jni-lib-musl /src/aws-lambda-runtime-interface-client.so > "${DST_DIR}"/classes/aws-lambda-runtime-interface-client.musl.so
+  mkdir "${DST_DIR}"/classes/$arch
+  docker build --platform=$platform -f "${SRC_DIR}/Dockerfile.glibc" --build-arg CURL_VERSION=${CURL_VERSION} -t lambda-java-jni-lib-glibc-$arch "${SRC_DIR}"
+  docker run --rm --entrypoint /bin/cat lambda-java-jni-lib-glibc-$arch /src/aws-lambda-runtime-interface-client.so > "${DST_DIR}"/classes/$arch/aws-lambda-runtime-interface-client.glibc.so
 
+  docker build --platform=$platform -f "${SRC_DIR}/Dockerfile.musl" --build-arg CURL_VERSION=${CURL_VERSION} -t lambda-java-jni-lib-musl-$arch "${SRC_DIR}"
+  docker run --rm --entrypoint /bin/cat lambda-java-jni-lib-musl-$arch /src/aws-lambda-runtime-interface-client.so > "${DST_DIR}"/classes/$arch/aws-lambda-runtime-interface-client.musl.so
+done
```
carlzogh, line 20: Variables suggested in https://github.com/carlzogh/aws-lambda-java-libs/commit/0b16ecde364fe03ea5036a9301581a1cfb9b9797 help keep our scripts more maintainable, inline variable mappings are not very easy to follow.

aws-lambda-java-runtime-interface-client/README.md
```diff
@@ -13,7 +13,7 @@ You can include this package in your preferred base image to make that base imag
 
 Choose a preferred base image. The Runtime Interface Client is tested on Amazon Linux, Alpine, Ubuntu, Debian, and CentOS. The requirements are that the image is:
 
-* built for x86_64
+* built for x86_64 and ARM
```
carlzogh, line 16: `ARM` -> `arm64`

carlzogh (changes requested):

aws-lambda-java-runtime-interface-client/README.md
muellerc: incorporated in the last update

aws-lambda-java-runtime-interface-client/src/main/jni/build-jni-lib.sh
muellerc: incorporated in the last update

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
muellerc: incorporated in the last update

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
muellerc: incorporated in the last update

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
```diff
@@ -13,15 +13,19 @@
  */
 class NativeClient {
     private static final String nativeLibPath = "/tmp/.aws-lambda-runtime-interface-client";
+    private static final String architecturePathSuffix = "/" + getArchIdentifier();
+    private static final String supported_arm_architectures = "^(aarch64)$";
+    private static final String supported_x86_architectures = "^(amd64)$";
```
carlzogh, line 18: This is still very restrictive and may not cater for all supported architectures - please change to the one proposed in my previously linked commit:

```java
        if (arch.matches("^(x8664|amd64|ia32e|em64t|x64|x86_64)$")) {
            return "x86_64";
        } else if (arch.startsWith("arm64") || arch.startsWith("aarch64")) {
            return "arm64";
        }
```

Also please include a comment describing where these patterns come from as suggested:

```java
// Implementation based on AWS CRT (ref. https://github.com/awslabs/aws-crt-java/blob/0e9c3db8b07258b57c2503cfc47c787ccef10670/src/main/java/software/amazon/awssdk/crt/CRT.java#L106-L134)
```

carlzogh (changes requested):

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/NativeClient.java
muellerc: I added the other architectures, to be less restrictive. But, I'm still using the regexp matches, so that we can still use a variable, holding all supported architectures. Hope this make sense to you.

carlzogh: That's cool, thanks @muellerc - could you please also add the comment I mentioned above with a link to the source of the regex/logic for future reference?

aws-lambda-java-events/pom.xml
```diff
@@ -159,6 +159,29 @@
...
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <version>3.8.1</version>
+            <configuration>
+              <annotationProcessorPaths>
+                <path>
+                  <groupId>org.projectlombok</groupId>
+                  <artifactId>lombok</artifactId>
+                  <version>${lombok.version}</version>
```
carlzogh, line 179: `lombok.version` should be defined under `<properties>` as:

```
<lombok.version>1.18.22</lombok.version>
```

aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/Dockerfile.agent
```diff
@@ -2,4 +2,5 @@ FROM public.ecr.aws/amazoncorretto/amazoncorretto:8
 
 RUN amazon-linux-extras enable docker && \
     yum clean metadata && \
-    yum install -y docker tar maven
+    yum install -y docker tar maven && \
+    alternatives --config java <<< '2'
```
carlzogh, line 6: Is this still required after changing the base image for this Dockerfile?

carlzogh (approved):

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/UnknownPlatformException.java
```diff
@@ -0,0 +1,11 @@
+package com.amazonaws.services.lambda.runtime.api.client.runtimeapi;
+
+/**
+ * Copyright (c) 2019 Amazon. All rights reserved.
```
smirnoal, line 4: 2022

smirnoal (approved):

smirnoal (approved):

zsombor-balogh (approved):

msailes:
I took @muellerc branch changed the project version number. Ran `make build` in the `aws-lambda-java-runtime-interface-client` directory to install the proposed changes into my local repo. 

Then I used this artefact to build a Lambda function which was then deployed to the ARM architecture on AWS Lambda. 

Whereas previously this would cause an exception, I'm now getting a successful response.

carlzogh:
Thanks @msailes!

carlzogh:
I've just released version `2.1.0` containing these changes - it should propagate to Maven Central and be available within a few hours. Thanks again @muellerc for your contribution and @msailes for your support!

davidmelia:
@carlzogh - have you seen https://github.com/aws/aws-lambda-java-libs/issues/316 - I think something has gone wrong with the JAR build as the ARM shared objects are actually x86

```
arm64/aws-lambda-runtime-interface-client.glibc.so: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, BuildID[sha1]=2b546f8d337d95e96fbf08f486a92ca29cde36da, not stripped
arm64/aws-lambda-runtime-interface-client.musl.so:  ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, with debug_info, not stripped
```

This means we cannot easily run on Apple M1 _(have to use x86 docker platform param   --platform=linux/amd64)_

Issue #224, opened 2021-02-17
DynamodbAttributeValueTransformer causing break in functionality for empty list when transforming (v2)


saemitang:
I'm using DynamoDb Enhanced library version `2.15.82`. Recently, `aws-lambda-java-events` version `3.7.0` causes a break in the functionality when trying to transform empty list. Empty list is transformed to `null` which for DynamoDb Enhanced library expects an empty list, and causes DynamoDB Enhanced library to throw the below stack trace.

```
java.lang.IllegalStateException: Unable to convert attribute value: AttributeValue()
software.amazon.awssdk.enhanced.dynamodb.internal.converter.attribute.EnhancedAttributeValue.fromAttributeValue(EnhancedAttributeValue.java:352) ~[dynamodb-enhanced-2.15.82.jar:na]
software.amazon.awssdk.enhanced.dynamodb.internal.converter.attribute.ListAttributeConverter$Delegate.transformTo(ListAttributeConverter.java:154) ~[dynamodb-enhanced-2.15.82.jar:na]
software.amazon.awssdk.enhanced.dynamodb.internal.converter.attribute.ListAttributeConverter.transformTo(ListAttributeConverter.java:120) ~[dynamodb-enhanced-2.15.82.jar:na]
software.amazon.awssdk.enhanced.dynamodb.internal.converter.attribute.ListAttributeConverter.transformTo(ListAttributeConverter.java:82) ~[dynamodb-enhanced-2.15.82.jar:na]
software.amazon.awssdk.enhanced.dynamodb.internal.mapper.StaticAttributeType.attributeValueToObject(StaticAttributeType.java:44) ~[dynamodb-enhanced-2.15.82.jar:na]
software.amazon.awssdk.enhanced.dynamodb.internal.mapper.ResolvedImmutableAttribute.lambda$create$1(ResolvedImmutableAttribute.java:61) ~[dynamodb-enhanced-2.15.82.jar:na]
software.amazon.awssdk.enhanced.dynamodb.mapper.StaticImmutableTableSchema.mapToItem(StaticImmutableTableSchema.java:458) ~[dynamodb-enhanced-2.15.82.jar:na]
software.amazon.awssdk.enhanced.dynamodb.mapper.WrappedTableSchema.mapToItem(WrappedTableSchema.java:54) ~[dynamodb-enhanced-2.15.82.jar:na]
```

Further investigation points to the below code in `DynamodbAttributeValueTransformer` that's causing the issue
```
[...]

PR #309, opened 2022-02-01
DynamodbAttributeValueTransformer v1 and v2 return empty list instead of null for empty list attribute

cayua:
*Issue #, if available:* https://github.com/aws/aws-lambda-java-libs/issues/224

*Description of changes:* 
This change will make DynamodbAttributeValueTransformer v1 and v2 return empty list instead of null for empty list input.
Transforming empty list to `null` will break functionality for DynamoDB Enhanced library  and make it throw runtime exceptions like:
```
java.lang.IllegalStateException: Unable to convert attribute value: AttributeValue()
```
Also updated corresponding unit test cases.


carlzogh (approved):
LGTM, thanks!

msailes (approved):
👍🏻

Issue #242, opened 2021-05-02
Any plan to add Lexv2 event?


scafarella:
Hi,

it seems that the new [lex v2 Input Event Format](https://docs.aws.amazon.com/lexv2/latest/dg/lambda.html#lambda-input-format) is different from the old LexEvent. 

Are there any plans to implement it?

Thanks

msailes:
@scafarella Would you be able to test the new event and response objects?

scafarella:
Thanks for the reply. Yes I can test it

Sordie:
@msailes is there already a Lex v2 event and response object somewhere? When I looked at the master branch I could only finde the Lex v1 event. If not, would you like me to open a pull request for both?

msailes:
@Sordie please do!

PR #321, opened 2022-04-04
Adding Lex v2 event and response

Sordie:
*Issue #, if available:*
#242

*Description of changes:*
Adding both Lex version 2 event and response as defined here: https://docs.aws.amazon.com/lexv2/latest/dg/lambda.html


aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LexV2Event.java
```diff
@@ -0,0 +1,228 @@
...
+
+    @Data
+    @Builder(setterPrefix = "with")
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class Transcription implements Serializable, Cloneable {
+        private String transcription;
+        private TranscriptionConfidence transcriptionConfidence;
+        private ResolvedContext resolvedContext;
+        private Map<String, Slot> ResolvedSlots;
```
msailes, line 210: is this a typo or does it only work with Capital 'R'?

msailes:
Could you add a test into the serialization module?

https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-serialization/src/test/java/com/amazonaws/services/lambda/runtime/serialization/events/LambdaEventSerializersTest.java

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LexV2Event.java
Sordie: Yes, must have missed that typo. Thanks for pointing it out.

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LexV2Event.java
```diff
@@ -0,0 +1,228 @@
...
+
+    @Data
+    @Builder(setterPrefix = "with")
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class SentimentScore implements Serializable, Cloneable {
+        private double mixed;
+        private double negative;
+        private double neutral;
+        private double positiv;
```
msailes, line 126: is this a typo?

msailes:
Hi @Sordie,

Could this PR be change to target the [V4](https://github.com/aws/aws-lambda-java-libs/tree/events-v4-serialization-v2) branch? You can use the Jackson annotations to handle the capitalisation.

msailes:
Could you check the tests?

Sordie:
Hi @msailes,
I have added an additional Lex event that includes the optional parts for both sentiment and Kendra. However, I’m still struggling with the tests. As already mentioned above the tests currently fail, because the slots map is empty after the deserialization and subsequent serialization. At this point I haven’t found an easy way to fix this. Could you share your opinion on how to best proceed? Maybe you already have an idea how to solve this issue?

msailes:
The `testLexV2Event` seems to be failing because the test input `lex_v2_event.json` contains a Map with null values. In the `JacksonPojoSerializer` class we set `mapper.setSerializationInclusion(Include.NON_NULL);` which excludes null values.

```
          "intent": {
              "slots": {
                  "ReturnDate": null,
                  "PickUpDate": null,
                  "DriverAge": null,
                  "CarType": null,
                  "PickUpCity": null
              },
```

I don't know if this is bad data or possible? @Sordie do you know?

msailes:
Same for the `testLexV2KendraSentimentEvent` test.

Sordie:
@msailes thanks for pointing out where the configuration to exclude null values is, as I had not seen it before.
Regarding the possibility of bad data in the test events, both event json files contain data taken directly from the input of a TypeScript Lambda that was attached to a Lex bot. I only replaced some unique identifiers with example strings. Therefore, I would say that this data is representative of the actual input a lambda will receive from the Lex service.

msailes:
@Sordie thank you.

I think the tests you've added should be slightly changed then to show this behavior. The actual will be different to the json from the test file because maps which include items will null values are excluded. 

@andclt, @smirnoal I'm not sure how you want this implementing, maybe two files would be a suitable way?

@andclt, @smirnoal I found this a hard problem to track down since there was no specific JSON knowledge in the testing lib. I would suggest with move to [JSONassert](https://github.com/skyscreamer/JSONassert) either in this PR or another seperate one. This library gave much better error messages and I was able to spot the problem straight away.

andclt:
First of all thanks for contributing @Sordie! 

I agree with @msailes' suggestion. However, if you want to keep those null values after the serialization, I think you can achieve it adding the `@JsonInclude` annotation (https://fasterxml.github.io/jackson-annotations/javadoc/2.9/com/fasterxml/jackson/annotation/JsonInclude.html) at the field level. This will override the global custom serialization behaviour that is currently excluding null values.

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LexV2Event.java
```diff
@@ -0,0 +1,288 @@
...
+ *
+ * @see <a href=
+ *      "https://docs.aws.amazon.com/lexv2/latest/dg/lambda.html#lambda-input-format">Using
+ *      an AWS Lambda function</a>
+ */
+@Data
+@Builder(setterPrefix = "with")
+@NoArgsConstructor
+@AllArgsConstructor
+public class LexV2Event implements Serializable, Cloneable {
```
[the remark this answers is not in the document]
Sordie: Ok, I have removed the cloneable interface from all classes I created

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LexV2Event.java
```diff
@@ -0,0 +1,288 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
Sordie: Makes sense, I have updated the year to 2022 :D

andclt:
Hi @Sordie,

As I said in a previous message on this thread, I think we should use the `@JsonInclude` annotation to override our current  Jackson configuration (that's excluding null values) instead of using a different json file to compare the lex event v2 serialization.

I was able to test successfully the following implementation, could you please take a look at it?

```
private Map<String, Slot> slots;

@JsonInclude(value=JsonInclude.Include.NON_NULL, content=JsonInclude.Include.ALWAYS)
public Map<String, Slot> getSlots() {
    return this.slots;
}

@JsonInclude(value=JsonInclude.Include.NON_NULL, content=JsonInclude.Include.ALWAYS)
public void setSlots(Map<String, Slot> slots) {
    this.slots = slots;
}
```

Sordie:
Hi @andclt, thanks for providing a code snippet. I will add your suggestion next week once I have a computer in front of me again.

Sordie:
Hi @andclt, I have applied your suggestion and removed the now unused json files.

andclt:
Hi @Sordie, thank you! Approved!

andclt (approved):

Issue #327, opened 2022-04-15
Uniformize Maven Projects via Parent


driverpt:

smirnoal:
Hey Luís, thanks for the suggestion. Could you please add a brief description of how it helps? what would be the benefit of it?

driverpt:
Ok, so what i'm trying to achieve is to add a Parent POM (also to work as BOM) in which we add `dependencyManagement` to have all the necessary dependencies without the need to explicitly add `<version>` to each of the projects

smirnoal:
The libraries in this repo are not very much coupled with one another. The intention here is to allow them evolve and be released independently, without much of cohesion. I believe the best option for you would be to maintain your local BOM.

We keep README.md in the root updated with the latest versions, please check to see if it helps

driverpt:
Then if they are not coupled, i suggest splitting them into multiple Repos. I recommend using a Parent POM to uniformize Libraries if they are going to stay in the same repo

smirnoal:
right, as it grows in future, I think it will end up with separate repos. Now it looks like a right balance to manage them all in a single monorepo

PR #328, opened 2022-04-15
Uniformize Maven Project

driverpt:
Addresses #327 

This PR attempts to add Dependency Management across all projects to prevent duplicated code


msailes:
No current plan to move to a parent pom.

PR #331, opened 2022-04-22
Events v4 serialization v2

andclt:
*Issue #, if available:*
https://github.com/aws/aws-lambda-java-libs/issues/282

*Description of changes:*
Initial implementation of `Events V4` and `Serialization V2`


smirnoal (approved):

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/IamPolicyResponse.java
```diff
@@ -1,15 +1,28 @@
...
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
 import java.io.Serializable;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
driverpt, line 25: No `*` imports please

aws-lambda-java-events/pom.xml
```diff
@@ -1,11 +1,11 @@
...
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/maven-v4_0_0.xsd">
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.amazonaws</groupId>
   <artifactId>aws-lambda-java-events</artifactId>
-  <version>3.11.0</version>
+  <version>4.0.0</version>
```
driverpt, line 8: Suggestion: Bump to Java 11 on v4

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/api_gateway/APIGatewayCustomAuthorizerEvent.java
```diff
@@ -1,4 +1,17 @@
...
+ * the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+
+package com.amazonaws.services.lambda.runtime.events.api_gateway;
```
driverpt, line 14: Suggestion: No `_` in Package Name

aws-lambda-java-events/src/test/java/com/amazonaws/services/lambda/runtime/events/models/s3/S3EventNotificationTest.java
```diff
@@ -27,9 +30,9 @@ public class S3EventNotificationTest {
     public void testGetUrlDecodedKey() {
         for (String testKey : KEYS_REQUIRING_URL_ENCODE) {
             String urlEncoded = HttpUtils.urlEncode(testKey, false);
-            S3EventNotification.S3ObjectEntity entity = new S3EventNotification.S3ObjectEntity(
-                    urlEncoded, 1L, "E-Tag", "versionId");
+            S3Event.S3ObjectEntity entity = new S3Event.S3ObjectEntity(
```
driverpt, line 33: Add `@Builder` ?

aws-lambda-java-serialization/pom.xml
```diff
@@ -31,47 +32,49 @@
     <properties>
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
-        <relocation.prefix>com.amazonaws.lambda.thirdparty</relocation.prefix>
+        <jackson.version>2.12.6.1</jackson.version>
```
driverpt, line 35: Suggestion: Create Parent POM with `dependencyManagement`. To act as BOM

aws-lambda-java-serialization/pom.xml
```diff
@@ -31,47 +32,49 @@
...
-            <version>2.8.5</version>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+            <version>${jackson.jsr10.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.json</groupId>
-            <artifactId>json</artifactId>
-            <version>20160810</version>
+            <groupId>com.google.code.gson</groupId>
```
driverpt, line 53: Why GSON and Jackson ?

aws-lambda-java-serialization/src/main/resources/META-INF/services/com.amazonaws.services.lambda.serialization.CustomPojoSerializer
```diff
@@ -0,0 +1 @@
+com.amazonaws.services.lambda.serialization.JacksonPojoSerializer
```
driverpt, line 1: How will translate when GraalVM is applied? Will it still use reflection?

aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/serialization/JacksonPojoSerializer.java
```diff
@@ -0,0 +1,222 @@
...
+        module.addDeserializer(Void.class, new VoidDeserializer());
+        mapper.registerModule(module);
+
+        JavaTimeModule javaTimeModule = new JavaTimeModule();
+        mapper.registerModule(javaTimeModule);
+
+        return mapper;
+    }
+
+    public static final class VoidDeserializer extends JsonDeserializer<Void> {
```
driverpt, line 135: What is this for ?

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/s3/S3Event.java
```diff
@@ -0,0 +1,185 @@
...
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class S3EventNotificationRecord {
+
+        private String awsRegion;
+        private String eventName;
+        private String eventSource;
+        // looking into removing jsonFormat when this issue will be resolved (https://github.com/FasterXML/jackson-modules-java8/issues/76)
+        @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSX")
+        private ZonedDateTime eventTime;
```
driverpt, line 167: Why not `Instant` ?

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/s3/S3Event.java
```diff
@@ -0,0 +1,185 @@
...
+        private String eventVersion;
+        private RequestParametersEntity requestParameters;
+        private ResponseElementsEntity responseElements;
+        private S3Entity s3;
+        private UserIdentityEntity userIdentity;
+
+    }
+
+    @Override
+    public S3Event clone() {
```
driverpt, line 177: Why would you want to clone this ?

aws-lambda-java-events/src/main/java/lombok.config
```diff
@@ -1 +0,0 @@
-lombok.getter.noIsPrefix = true
```
driverpt, line 1: Suggestion:
```
lombok.addLombokGeneratedAnnotation = true

```

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ConfigEvent.java
```diff
@@ -13,400 +13,84 @@
...
     /**
-     * Sets the ID of the AWS Config rule that triggered the event.
-     * @param configRuleId String containing the AWS Config rule ID.
+     * The JSON-encoded map containing the AWS Config rule parameters.
+     *
      */
-    public void setConfigRuleId(String configRuleId) {
-        this.configRuleId = configRuleId;
-    }
+    private String ruleParameters;
```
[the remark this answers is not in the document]
andclt: Because `ruleParameters` is a String. 

Ref: https://docs.aws.amazon.com/config/latest/developerguide/evaluate-config_develop-rules_example-events.html

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
```diff
@@ -36,20 +39,31 @@
 public class LambdaDestinationEvent implements Serializable, Cloneable {
 
     private String version;
-    private DateTime timestamp;
+
+    // looking into removing jsonFormat when this issue will be resolved (https://github.com/FasterXML/jackson-modules-java8/issues/76)
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSX")
+    private ZonedDateTime timestamp;
```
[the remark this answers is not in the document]
andclt: `timestamp` parameter requires Moment with time zone (`"2019-11-14T18:16:05.568Z"`), not moment in UTC 

Ref: https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html"

aws-lambda-java-events/pom.xml
```diff
@@ -46,30 +46,23 @@
...
+          <artifactId>gson</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
       <groupId>org.junit.jupiter</groupId>
       <artifactId>junit-jupiter-engine</artifactId>
-      <version>5.7.0</version>
+      <version>5.8.2</version>
```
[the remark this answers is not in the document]
andclt: Good catch, thanks!

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/IamPolicyResponse.java
andclt: Thanks!

aws-lambda-java-events/pom.xml
andclt: We still support running this library with java 8 so we should still be targeting it.

aws-lambda-java-serialization/pom.xml
andclt: Conversation going on here https://github.com/aws/aws-lambda-java-libs/issues/327

aws-lambda-java-serialization/pom.xml
andclt: `GSON` is still used by the [RIC](https://github.com/aws/aws-lambda-java-libs/blob/master/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/EventHandlerLoader.java#L82) during serialization when the client context platform is Android

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/s3/S3Event.java
andclt: Same reason as `LambdaDestinationEvent`

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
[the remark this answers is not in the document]
driverpt: `UTC` = `Z`

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/s3/S3Event.java
andclt: Simply not to break customers functions if they are relying on it because the `clone` method was present in v3 as well.

aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/serialization/JacksonPojoSerializer.java
andclt: In order to serialize `Void` as not `null`.

driverpt: Can you give me an example ?

aws-lambda-java-serialization/pom.xml
driverpt: Why not migrate Android to Jackson ?

aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/serialization/JacksonPojoSerializer.java
andclt: Sure, a specific test for it was created https://github.com/aws/aws-lambda-java-libs/blob/c1685c79742aae77609d9dd1beb38f214a0664b1/aws-lambda-java-serialization/src/test/java/com/amazonaws/services/lambda/serialization/JacksonPojoSerializerTest.java

aws-lambda-java-events/src/test/java/com/amazonaws/services/lambda/runtime/events/models/s3/S3EventNotificationTest.java
andclt: Done - thanks

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/api_gateway/APIGatewayCustomAuthorizerEvent.java
andclt: Renamed it to `apigateway` - thanks.

aws-lambda-java-serialization/src/main/java/com/amazonaws/services/lambda/serialization/JacksonPojoSerializer.java
driverpt: I'm still trying to understand why you need the `Void` type. Is there any real-world example?

aws-lambda-java-serialization/pom.xml
andclt: It would be a breaking change for customers that are using the Java runtime for developing Android applications because the Jackson's serialization for Optionals, Date, LocalDate and Calendar types is different compared to the Gson one.

However, we are exploring this opportunity.

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/LambdaDestinationEvent.java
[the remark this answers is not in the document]
andclt: You're right I'm sorry. What I wanted to say is that right now we have been using Joda `DateTime` for timestamps which is TZ-aware, so `ZonedDateTime` was the correct replacement. 

However, since we're defining timestamps in UTC/GMT, we're fine using `Instant`.

aws-lambda-java-serialization/src/main/resources/META-INF/services/com.amazonaws.services.lambda.serialization.CustomPojoSerializer
andclt: From [1.0-RC8](https://www.graalvm.org/release-notes/release-candidates/#10-rc8):
> The native image generator now has automatic support for services loaded using ServiceLoader. All service implementation classes, listed in the META-INF directory, are available automatically as soon as the service interface is used. This eliminates the need to manually register resources and reflection support for such classes. The automatic registration can be disabled with the -H:-UseServiceLoaderFeature option.

aws-lambda-java-serialization/pom.xml
driverpt: That's why v4 is a Major Version. Assumes there's breaking changes.

AdamBien:
I reviewed the recent changes / diffs. Built-in Java libraries, like e.g. java.time reduce the complexity and make maintenance easier. 
Lombok is controversial - essentially, lombok is a code generator, so the coupling is tighter and the dependency more critical.
I prefer JSON-B / JSON-P for json handling - however, Jackson can be also considered as "quasi" standard.

GoodforGod:
1) Standardization is definitely what was missing and its good that it is coming in v4. Nice that code base become consistent.

2) Lombok in this type of library in my opinion is okay (only Java 11+), cause the whole purpose of library is to provide DTO, lombok allow for easier source code reading for developers and allow devs to maintain library easier, win-win.

3) Ability to instantiate it via new Event(), new Event(constructor ... args), lombok builder, static builders, DTOs are mutable even there is builder.. kinda messed up for DTO library, too many API for the same purpose.

Data annotation with Accessors(chain = true) is more than enough in my opinion.

4) Jackson as a first citizen serializer is not a good way for a DTO library. Cause now lib is incompatible with other serializers at this point (GSON for example). It should be then aws-jackson-events library to point directly that it is jackson focused lib or support all types of serializers by proper field naming.

5) Removing joda-time is a way to go, should be done like starting when Java 8 was out, but thats better late than never. In my opinion it is better to serialize time into ZonedDateTime and not Instant with force UTC time conversion, cause that's not what users expects.

6) Keeping clone interface compatability for v4 is not the way to go in my opinion, v4 is a Major version so thats okay if there are breaking changes (devs should expect that), so clone methods should be removed.

7) Some events when created initialize new collections, like new ArrayList(), collection will prob be fully replaced by serializer so this just generates garbage at this point, can be cleaner up.

Here is some feedback and my thoughts on this topic 🙃

PR #338, opened 2022-05-06
Fix os compatibility test local builds on arm64 hosts

zsombor-balogh:
*Issue #, if available:*

*Description of changes:*
Fix os compatibility tests on arm64 host local builds:
* update `codebuild_build.sh` script from upstream to be able to set `local_agent_image`
* use `arm64` specific local agent image when running on `arm64` host
* extract invocation test logic to a separate script
* increase invocation timeout when running test on `arm64` with `x86_64` being emulated (this combination timed out with previous 10 sec configuration)
* set platform explicitly when running test images to avoid architecture discrepancy warning message

### DIFF
--- a/aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/codebuild_build.sh
+++ b/aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/codebuild_build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# This file is copied from https://github.com/aws/aws-codebuild-docker-images/blob/f0912e4b16e427da35351fc102f0f56f4ceb938a/local_builds/codebuild_build.sh
+# This file is copied from https://github.com/aws/aws-codebuild-docker-images/blob/282c6634e8c83c2a9841719b09aabfced3461981/local_builds/codebuild_build.sh
 
 function allOSRealPath() {
     if isOSWindows
@@ -36,6 +36,7 @@ function usage {
     echo "  -a        Used to specify an artifact output directory."
     echo "Options:"
     echo "  -l IMAGE  Used to override the default local agent image."
+    echo "  -r        Used to specify a report output directory."
     echo "  -s        Used to specify source information. Defaults to the current working directory for primary source."
     echo "               * First (-s) is for primary source"
     echo "               * Use additional (-s) in <sourceIdentifier>:<sourceLocation> format for secondary source"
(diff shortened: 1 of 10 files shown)


aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/codebuild_build.sh
```diff
@@ -95,7 +97,7 @@ then
     exit 1
 fi
 
-docker_command="docker run "
+docker_command="docker run -it "
```
carlzogh, line 100: I think interactive mode only works with TTY, I believe this is the cause of the checks failing - any reason we need `-it` here?

aws-lambda-java-runtime-interface-client/test/integration/codebuild/buildspec.os.alpine.yml
```diff
@@ -79,34 +79,7 @@ phases:
...
-      - |
-        echo "Response: ${actual}"
-        if [[ "$actual" != "$expected" ]]; then
-          echo "fail! runtime: $RUNTIME - expected output $expected - got $actual"
-          exit -1
-        fi
+      - aws-lambda-java-runtime-interface-client/test/integration/codebuild/scripts/run_invocation_test.sh
     finally:
       - |
         echo "---------Container Logs: ${OS_DISTRIBUTION}-app----------"
```
carlzogh, line 85: nit: could also extract the cleanup logic to a separate script?

aws-lambda-java-runtime-interface-client/test/integration/codebuild/scripts/run_invocation_test.sh
```diff
@@ -0,0 +1,41 @@
...
+docker network create "${OS_DISTRIBUTION}-network"
+
+docker run \
+  --detach \
+  --name "${OS_DISTRIBUTION}-app" \
+  --network "${OS_DISTRIBUTION}-network" \
+  --entrypoint="" \
+  --platform="${PLATFORM}" \
+  "${IMAGE_TAG}" \
+  sh -c "/usr/bin/${RIE} ${JAVA_BINARY_LOCATION} -jar ./HelloWorld-1.0.jar helloworld.App"
```
carlzogh, line 16: nit: extract jar and handler to their own variables for better readability

aws-lambda-java-runtime-interface-client/test/integration/codebuild/scripts/run_invocation_test.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
```
carlzogh, line 1: Thanks for extracting this out! Much neater this way :)

aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/codebuild_build.sh
zsombor-balogh: Actually no we don't, it was just copied from upstream as it is. I will remove it.

aws-lambda-java-runtime-interface-client/test/integration/codebuild/buildspec.os.alpine.yml
zsombor-balogh: I wanted to extract only in a single commit/PR what I "touch", but they logically belong together it's reasonable to extract them in the same PR, will do.

carlzogh: Thanks for doing this!

smirnoal (approved):

PR #367, opened 2022-08-30
1.2.2 lambda-java-core release

andclt:
*Description of changes:*
### August 30, 2022
`1.2.2`:
- Added new `CustomPojoSerializer` interface
- Removed unnecessary usage of public on interface methods (aws#172) 

### DIFF
--- a/aws-lambda-java-core/src/main/java/com/amazonaws/services/lambda/runtime/CustomPojoSerializer.java
+++ b/aws-lambda-java-core/src/main/java/com/amazonaws/services/lambda/runtime/CustomPojoSerializer.java
@@ -1,0 +1,14 @@
+/* Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved. */
+
+package com.amazonaws.services.lambda.runtime;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+
+import java.lang.reflect.Type;
+
+public interface CustomPojoSerializer {
+    <T> T fromJson(InputStream input, Type type);
+    <T> T fromJson(String input, Type type);
+    <T> void toJson(T value, OutputStream output, Type type);
+} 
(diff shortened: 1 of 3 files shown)


smirnoal (approved):

richarddd:
Wouldn't it also make sense to have a 4th method in the CustomSerializer:
`<T> String toJson(T value, Type type);`
Then the same serializer instance can be used inside handlers to create event responses requiring json strings

andclt:
> Wouldn't it also make sense to have a 4th method in the CustomSerializer: `<T> String toJson(T value, Type type);` Then the same serializer instance can be used inside handlers to create event responses requiring json strings

As discussed offline, we do not want to force cx to implement that method when creating a custom serializer. Furthermore, if they want, they can always add it to their custom serializer class.

aws-lambda-java-core/src/main/java/com/amazonaws/services/lambda/runtime/CustomPojoSerializer.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Interface required to implement a custom plain old java objects serializer
+ */
+public interface CustomPojoSerializer {
+
+    /**
+     * Deserializes from input stream to plain old java object
+     * @param input input stream
+     * @param type plain old java object type
+     * @return deserialized plain old java object of type T
```
pzygielo, line 19: What's `T`?

andclt: It's the type of the deserialized instance

pzygielo: I see the intention. But "deserialized plain old java object of type T" uses `T` which is undefined and unused anywhere else.

msailes (approved):
LGTM

PR #369, opened 2022-10-08
[RFC] Switch to pure HttpUrlConnection Implementation

driverpt:


msailes:
Hi Luís,

Do you have any performance data on this change?

Thanks,

driverpt:
Not yet @msailes , but the purpose is to make this Client OS/Arch agnostic

driverpt:
@msailes , comparing smoke tests output, with `HttpURLConnection` the invocations are slightly faster. Don't forget that JNI calls are typically expensive because they are not "JIT-able" or Optimizable by the JVM.

How do you suggest to Benchmark this properly with nice reports?

marksailes:
To start with I would test a few thousand requests with a minimal handler (hello world) with the java 11 managed runtime. Then do the same thing again with your 'custom' Java runtime.

richarddd:
> @msailes , comparing smoke tests output, with `HttpURLConnection` the invocations look slightly faster. Don't forget that JNI calls are typically expensive because they are not "JIT-able" or Optimizable by the JVM. It's possible to optimize even further the current code (Keep Alive, etc...)
> 
> How do you suggest to Benchmark this properly with nice reports?

Hey @driverpt, Richard from AWS here. Thanks for the effort in creating this PR! Much appreciated 🙌

The reason why we have a native implementation is purely for performance (and more importantly startup performance). We have seen that using a native c++ client greatly reduces cold starts because there is no code to JIT by the C1 or C2 compilers. HttpURLConnection is a rather old implementation with support for a lot of deprecated protocols and could not be replaced by [HttpClient](https://openjdk.org/groups/net/httpclient/intro.html). But even HttpClient will add some latency to first invokes compared to the native implementation.

The most ideal way to go about this is to let users opt-in and choose their own http client via a ServiceLoader interface.

I would be happy to take a look at such a PR if you're interested :)

Again thanks for your effort!

(BTW [HttpURLConnection uses a connection pool by default](https://docs.oracle.com/javase/7/docs/technotes/guides/net/http-keepalive.html) unless you opt-out)

driverpt:
Hello @richarddd , thanks for your reply.

I think JNI calls should only be used as a last resort, since they add a lot of complexity to the code base. I do believe that you will gain a couple of milliseconds in terms of coldstart but it's not optimal because you will lose the Java Heap Management.

I like to tinker with Java in terms of performance and I strongly believe that it's possible to optimize this even further like those opt-outs you mentioned and to use Threads to initialize the Client while performing class/method lookups for the handler.

Java is a fast language, it's just a matter of tinkering with the JIT to make the code faster.

richarddd:
> Hello @richarddd , thanks for your reply.
> 
> I think JNI calls should only be used as a last resort, since they add a lot of complexity to the code base. I do believe that you will gain a couple of milliseconds in terms of coldstart but it's not optimal because you will lose the Java Heap Management.

From what we've seen its more than 100+ milliseconds difference inside Lambda :(

> I like to tinker with Java in terms of performance and I strongly believe that it's possible to optimize this even further like those opt-outs you mentioned and to use Threads to initialize the Client while performing class/method lookups for the handler.
> 
> Java is a fast language, it's just a matter of tinkering with the JIT to make the code faster.

Yes, Java is fast, it's not just fast at startup. Most customers will benefit from having an HTTP client with less peak performance over one that's fast once warmed up (achieved by JITed code) in a Serverless environment such as Java.

If you're measurements are different I would be very interested to see your numbers of this deployed inside Lambda :)

smirnoal:
thanks @driverpt for the contribution. I was also trying different http client implementation in my personal fork some time ago: https://github.com/smirnoal/aws-lambda-java-libs/branches 

Please do not close this PR for now. I think we can benefit from this implementation for Java8. I will be separating out http client interfaces and implementation into a dedicated package, which would be consumed by RIC(Runtime Interface Client), I have some ideas around this.

PR #373, opened 2022-11-09
Added properties field to AMQEvent.AMQMessage

lohithpc1:
*Issue #, if available:*
Custom Properties Set by the customer are not being propagated to target function, as we don't have a properties field in here.

*Description of changes:*
Added properties field and also a unit test to see if we are getting right fields.


smirnoal (dismissed):

smirnoal:
I think this should be merged to `events-v4-serialization-v2`, but not to `new_branch`

toniStepanofAWS (dismissed):

PR #394, opened 2023-01-08
Adding reference to calling class for reflection issue in unit tests.

tshortess:
*Issue #, if available:* Although I have not raised an issue for this, if what I describe is reproducible by others, I propose the changes in this pull request to resolve the issue. I noticed this evening when I started writing unit tests that the HandlerParamsArguments option was not working with either just events or events and responses combined as I continued to get the "Path /resources/events/ cannot be found" IllegalArgumentException. I tried all variations of passing in the folder location. To test that my path was correct, I created a test program using the same methods and logic as the methods being called, which confirmed that the issue was not with the path string. I then walked through the debugger in my unit test and it showed that the class that was considered the caller and passed to the getResource and getResourceAsStream methods was the AWS class. I would note that the only reason the single event loader works is because of the "stream = new FileInputStream(new File(filename));" line - the other first two attempts to set the input stream likewise return null.  All unit tests pass following my changes, and if you believe a better solution exists, I'm all for it. I just want to enable the multi-events and responses feature in my unit tests. Thanks!

*Description of changes:* The changes in this PR 1) add a reference to the calling class to obtain files and folders and 2) remove the extraneous attempts to get the file or folder path that always resulted in null.

### DIFF
--- a/aws-lambda-java-tests/src/main/java/com/amazonaws/services/lambda/runtime/tests/EventLoader.java
+++ b/aws-lambda-java-tests/src/main/java/com/amazonaws/services/lambda/runtime/tests/EventLoader.java
@@ -117,10 +117,7 @@ public class EventLoader {
 
         PojoSerializer<T> serializer = LambdaEventSerializers.serializerFor(targetClass, ClassLoader.getSystemClassLoader());
 
-        InputStream stream = serializer.getClass().getResourceAsStream(filename);
-        if (stream == null) {
-            stream = serializer.getClass().getClassLoader().getResourceAsStream(filename);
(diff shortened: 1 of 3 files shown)


msailes:
Do you have a project which reproduces the problem? 

Could you add unit tests to the PR?

tshortess:
I'm going to close this PR. I figured out the solution to my issue while trying to reproduce in a new project, which I've pushed [here](https://github.com/tshortess/example-java-lambda-testing). Although I had a folder named "resource" in my test path, it was not getting treated as the resource folder. My apologies, Mark.

PR #410, opened 2023-03-10
1.1.2 aws-lambda-java-serialization release

smirnoal:
*Description of changes:*
* Update `jackson-databind` from `2.10.5.1` to `2.14.2`
* Update `gson` from `2.8.9` to `2.10.1`
* Update `json` from `20160810` to `20230227`
* Register JDK8, JavaTime and Joda modules for jackson-serialization.
* Remove unit tests inside the `aws-lambda-java-serialization` lib and run the test suite from `aws-lambda-java-tests`.
* Remove VoidSerializer used to serialize void as not null
* Stage `1.1.2` serialization lib release

**Note**:
1. `jackson-databind >= v2.12.0` explicitly fails (de)serialization of `org.joda.time.*` types in absence of registered custom (de)serializers. [PR #2776](https://github.com/FasterXML/jackson-databind/issues/2776). Registering `JodaModule` module is required to (de)serialize `org.joda.time.*` types. However, the `JodaModule` brings the joda time dependency which we need to relocate in order not to clash with the one provided by customers (if they provide one). Due to the following [MSHADE-156 bug](https://issues.apache.org/jira/browse/MSHADE-156), the Maven shade plugin will relocate string literals too, breaking our serialization logic. As a workaround, adding a 'fake' relocation pattern in order to prevent such behaviour. 
2. Unit tests were removed because they do not run against the shaded jar so they fail. `aws-lambda-java-tests` package contains a more exhaustive test suite so the Github workflow will now run those tests when editing `aws-lambda-java-serialization` lib.
3. VoidSerializer is not working in Java17 because after Java11 there was a change in the implementation of the `java.lang.Void class`. In java11 the Void class had a no-arg constructor that was accessible, which allowed us to create a new instance of Void using reflection but starting from Java12 the Void class was modified to have a private constructor that takes no arguments in order to prevent its instantiation.


aws-lambda-java-runtime-interface-client/pom.xml
```diff
@@ -56,7 +56,7 @@
     <dependency>
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-lambda-java-serialization</artifactId>
-      <version>1.0.0</version>
+      <version>1.1.0</version>
```
andclt, line 59: After releasing the new serialization-lib we should update the dependency here

andclt (approved):

aws-lambda-java-runtime-interface-client/pom.xml
smirnoal: yes, will do a separate PR once the library is published to maven

anton-stepanof (approved):

PR #442, opened 2023-09-18
Objects to represent VPC Lattice V2 request and response.

msailes:
*Issue #, if available:*
n/a

*Description of changes:*

Objects to represent VPC Lattice V2 request and response as documented https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html

*Target (OCI, Managed Runtime, both):* n/a

### DIFF
--- a/aws-lambda-java-events/pom.xml
+++ b/aws-lambda-java-events/pom.xml
@@ -50,6 +50,11 @@
             <artifactId>joda-time</artifactId>
             <version>2.10.8</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+            <version>3.0.2</version>
+        </dependency>
 
         <dependency>
             <groupId>org.junit.jupiter</groupId>
--- a/aws-lambda-java-events/pom.xml
+++ b/aws-lambda-java-events/pom.xml
@@ -51,9 +51,9 @@
             <version>2.10.8</version>
         </dependency>
         <dependency>
-            <groupId>com.google.code.findbugs</groupId>
-            <artifactId>jsr305</artifactId>
-            <version>3.0.2</version>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>24.0.1</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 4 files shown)


deki:
Great just created #443 for it (forgot to hit submit yesterday).

As per https://docs.aws.amazon.com/lambda/latest/dg/services-vpc-lattice.html#vpc-lattice-receiving-events `x509SanDns` should be added as well.

msailes:
Added

aws-lambda-java-events/pom.xml
```diff
@@ -50,6 +50,11 @@
             <artifactId>joda-time</artifactId>
             <version>2.10.8</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
```
deki, line 55: we shouldn't add that dependency for everyone

aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/VpcLatticeV2RequestEvent.java
```diff
@@ -0,0 +1,72 @@
...
+        private Identity identity;
+        private String region;
+        /**
+         * Number of microseconds from the epoch
+         */
+        private String timeEpoch;
+
+        public LocalDateTime getlocalDateTime() {
+            // Truncating to milliseconds. Java 8 only supports Epoch with precision of milliseconds. Microseconds are supported in Java 9
+            long epochMilli = Long.parseLong(timeEpoch.substring(0, timeEpoch.length() - 3));
```
deki, line 56: I'd add a null check even if the field is supposed to be always present. In addition provide a better method name for `getlocalDateTime`.

smirnoal: if it is null, we'll get NPE in this line - and fail fast. I think this is expected behaviour in that case, as there's no reasonable fallback.

smirnoal: On another note, I'd avoid truncating string in this line
my suggestion would be
```
long epochMicroseconds = Long.parseLong(timeEpoch);
long epochMilliseconds = epochMicroseconds / 1000;

return Instant.ofEpochMilli(epochMilliseconds).atZone(ZoneOffset.UTC).toLocalDateTime();
```

msailes: Done

aws-lambda-java-events/pom.xml
msailes: I've changed the dependency to the jetbrains one. There is no standard for this yet, which is a shame. I think it's important to describe the data of the events as best as we can. I hope that ```@Nullable``` will be an annotation that is added to more events in the future.

sdelamo:
would it be possible to have Json samples of this event? it will allow us to test serialization.

PR #449, opened 2023-10-17
Change AwsProxyResponseEvent's isBase64Encoded field to boolean primitive type to avoid NullPointer…

mbfreder:
…Exception

*Issue #, if available:*

*Description of changes:*

Change AwsProxyResponseEvent's isBase64Encoded field to boolean primitive type to avoid NullPointer

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/AwsProxyResponseEvent.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/AwsProxyResponseEvent.java
@@ -29,5 +29,5 @@ public class AwsProxyResponseEvent {
 
     private String body;
 
-    private Boolean isBase64Encoded;
+    private boolean isBase64Encoded;
 }


smirnoal:
will it be false by default? do we have a serialization test for this case?

mbfreder:
Yes, it will be false by default. [`APIGatewayV2HTTPResponse`](https://github.com/aws/aws-lambda-java-libs/blob/d5930f1d65d0e25898d7b9a9e6a7337c7c41ebad/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/apigateway/APIGatewayV2HTTPResponse.java#L35C37-L35C37) already uses boolean primitive type for that field and there's already a serialization test for that [here](https://github.com/aws/aws-lambda-java-libs/blob/d5930f1d65d0e25898d7b9a9e6a7337c7c41ebad/aws-lambda-java-events/src/test/java/com/amazonaws/services/lambda/runtime/events/LambdaEventSerializersTest.java#L88).

PR #456, opened 2023-10-26
change scope for jetbrains annotations to provided

deki:
*Description of changes:*
We should avoid we introduce a particular annotation library into the dependency of users of the library. There are a lot of libraries out there and users may have a different preference. I've changed the scope to provided according to the docs https://github.com/JetBrains/java-annotations#using-the-annotations.

### DIFF
--- a/aws-lambda-java-events/pom.xml
+++ b/aws-lambda-java-events/pom.xml
@@ -61,6 +61,7 @@
       <groupId>org.jetbrains</groupId>
       <artifactId>annotations</artifactId>
       <version>24.0.1</version>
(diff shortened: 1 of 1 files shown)


msailes:
Instead of the annotations we could have these attribute be Optional<Type>. What do you think about that instead?

deki:
Not sure https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555

smirnoal (approved):

smirnoal:
thanks Dennis

PR #482, opened 2024-06-10
Fixing Queue representation in ConnectEvent.java

jonnysw:
*Issue #, if available:*
NA

*Description of changes:*
At present, when trying to invoke a lambda that uses a ConnectEvent in its handler, it will fail with:
```
An error occurred during JSON parsing: java.lang.RuntimeException
java.lang.RuntimeException: An error occurred during JSON parsing
Caused by: java.io.UncheckedIOException: com.amazonaws.lambda.thirdparty.com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
 at [Source: (ByteArrayInputStream); line: 1, column: 581] (through reference chain: com.amazonaws.services.lambda.runtime.events.ConnectEvent["Details"]->com.amazonaws.services.lambda.runtime.events.ConnectEvent$Details["ContactData"]->com.amazonaws.services.lambda.runtime.events.ConnectEvent$ContactData["Queue"])
```

I am fixing the Event model so that it can be correctly deserialised.

AWS Connect defines the Queue as an object when invoking a lambda, not a string. The object contains ARN and Name (as well as Address and Type). See here for the full model:
https://docs.aws.amazon.com/connect/latest/adminguide/connect-lambda-functions.html

For reference, I have followed the model used by the Go implementation here: https://github.com/aws/aws-lambda-go/blame/main/events/connect.go

both

### DIFF
--- a/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ConnectEvent.java
+++ b/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/ConnectEvent.java
@@ -59,7 +59,7 @@ public class ConnectEvent implements Serializable, Cloneable {
         private String initiationMethod;
         private String instanceArn;
         private String previousContactId;
-        private String queue;
+        private Queue queue;
         private SystemEndpoint systemEndpoint;
     }
 
@@ -72,6 +72,15 @@ public class ConnectEvent implements Serializable, Cloneable {
         private String type;
     }
 
+    @Data
+    @Builder(setterPrefix = "with")
+    @NoArgsConstructor
+    @AllArgsConstructor
+    public static class Queue implements Serializable, Cloneable {
+        private String name;
+        private String arn;
+    }
+    
     @Data
     @Builder(setterPrefix = "with")
     @NoArgsConstructor


smirnoal:
would you mind updating https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-tests/src/test/resources/connect_event.json accordingly?

jonnysw:
Thanks for taking a look @smirnoal - apologies for missing the tests, I'd just opened that one file in the online editor and hadn't thought to check. I've updated accordingly now, appreciate if you could take another look.

msailes:
While we're changing the event, I think it would be wise to include all the changes to Queue.

```
            "Queue": {
                   "ARN": "arn:aws:connect:eu-west-2:111111111111:instance/cccccccc-bbbb-dddd-eeee-ffffffffffff/queue/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee",
                 "Name": "PasswordReset"
                "OutboundCallerId": {
                    "Address": "+12345678903",
                    "Type": "TELEPHONE_NUMBER"
                }
            },
```

smirnoal:
@jonnysw would you take a look at the failed test?

joysl:
Hi Team,
Customer would like this to be fixed, please lmk if it's something I can help with to pass the CI flow

PR #506, opened 2024-08-09
Fix name of s3Bucket field of Task class in S3BatchEventV2

mmerfort:
*Description of changes:*

Fix the name of the s3Bucket field, which was previously incorrect. Add a test for that specific field.

unknown


maxday (approved):
LGTM!

smirnoal:
can you please share a link to public doc with the correct name?

mmerfort:
Documentation on this topic seems a bit sparse and this is the only thing I could find: https://aws.amazon.com/blogs/storage/automate-object-processing-in-amazon-s3-directory-buckets-with-s3-batch-operations-and-aws-lambda/

smirnoal (approved):

Issue #516, opened 2024-12-09
Issue with Map types in CognitoUserPoolPreTokenGenerationEventV2 -> GroupOverrideDetails 


torbjokv:
According to my support representative (and maybe the documentation?) the response should return list types for groupsToOverride and iamRolesToOverride even when I use V2 pre token generation, not Map<String, String>.  My experiences with this is that cognito fails when returning an object here instead of a list also, so I suspect that the SDK uses the wrong types here. Thank you in advance for looking into my issue!

lachlanodonnell:
I found this out the hard way - after spending a couple of hours I eventually use Jackson to convert the input CognitoUserPoolPreTokenGenerationEventV2 object to a Map and manually set the List for response group override.  Please update the Java objects for CognitoUserPoolPreTokenGenerationEventV2.GroupOverrideDetails as they should be Lists, not Maps.

Found on 3.14.0 (maven central).

PR #517, opened 2024-12-09
Fix CognitoUserPoolPreTokenGenerationEventV2 group override details to list

torbjokv:
*Issue #516*

*Description of changes:*
CognitoUserPoolPreTokenGenerationEventV2 -> GroupOverrideDetails Response should return list types for groupsToOverride and iamRolesToOverride even when I use V2 pre token generation, not Map<String, String>

**This PR is mostly for me to explain the issue and will probably not build because of missing bumping of model version as I don't know how to do that**


lachlanodonnell:
I found this out the hard way - after spending a couple of hours I eventually use Jackson to convert the input CognitoUserPoolPreTokenGenerationEventV2 object to a Map and manually set the List for response group override. Please update the Java objects for CognitoUserPoolPreTokenGenerationEventV2.GroupOverrideDetails as they should be Lists, not Maps.

Found on 3.14.0 (maven central).

torbjokv:
solved by https://github.com/aws/aws-lambda-java-libs/pull/519

PR #531, opened 2025-03-18
Fix env vars override bug.

Talhumaidi:
Currently, the script overrides existing env vars. This is dangerous, especially if someone used it by mistake for production lambda.


maxday:
Good catch! Thanks for your contribution!

maxday (approved):

msailes (approved):
I validated the issue, and tested the fix. Looks good to me.

PR #540, opened 2025-04-16
Add tenant id to lambda context and structured log messages

ramisa2108:
*Issue #, if available:*

*Description of changes:*
- Add tenant id to LambdaContext object and structured log messages 

- Both

### DIFF
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/EventHandlerLoader.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/EventHandlerLoader.java
@@ -581,6 +581,7 @@ public final class EventHandlerLoader {
                         cognitoIdentity,
                         LambdaEnvironment.FUNCTION_VERSION,
                         request.getInvokedFunctionArn(),
+                        request.getTenantId(),
                         clientContext
                 );
 
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/api/LambdaContext.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/api/LambdaContext.java
@@ -22,6 +22,7 @@ public class LambdaContext implements Context {
     private final long deadlineTimeInMs;
     private final CognitoIdentity cognitoIdentity;
     private final ClientContext clientContext;
+    private final String tenantId;
     private final LambdaLogger logger;
 
     public LambdaContext(
@@ -34,6 +35,7 @@ public class LambdaContext implements Context {
             CognitoIdentity identity,
             String functionVersion,
             String invokedFunctionArn,
+            String tenantId,
             ClientContext clientContext
     ) {
         this.memoryLimit = memoryLimit;
@@ -46,6 +48,7 @@ public class LambdaContext implements Context {
         this.clientContext = clientContext;
         this.functionVersion = functionVersion;
         this.invokedFunctionArn = invokedFunctionArn;
+        this.tenantId = tenantId;
         this.logger = com.amazonaws.services.lambda.runtime.LambdaRuntime.getLogger();
     }
 
@@ -89,6 +92,10 @@ public class LambdaContext implements Context {
         long now = System.currentTimeMillis();
         int delta = (int) (this.deadlineTimeInMs - now);
         return delta > 0 ? delta : 0;
+    }
+
+    public String getTenantId() {
+        return tenantId;
     }
 
     public LambdaLogger getLogger() {
--- a/aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/include/aws/lambda-runtime/runtime.h
+++ b/aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/include/aws/lambda-runtime/runtime.h
@@ -62,6 +62,11 @@ struct invocation_request {
     std::chrono::time_point<std::chrono::system_clock> deadline;
 
     /**
+     * Tenant ID of the current invocation.
+     */
+    std::string tenant_id;
+
+    /**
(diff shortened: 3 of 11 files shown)


aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImplTest.java
```diff
@@ -333,6 +334,31 @@ public void nextTest() {
         }
     }
 
+    @Test
```
maxday, line 337: great test! 💯

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/EventHandlerLoader.java
```diff
@@ -581,7 +581,8 @@ public ByteArrayOutputStream call(InvocationRequest request) throws Error, Excep
                         cognitoIdentity,
                         LambdaEnvironment.FUNCTION_VERSION,
                         request.getInvokedFunctionArn(),
-                        clientContext
+                        clientContext,
```
maxday, line 584: I think it would be great to leave the clientContext as the last parameter? This way we keep aws-related fields grouped. WDYT?

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/api/LambdaContext.java
```diff
@@ -34,7 +35,8 @@ public LambdaContext(
             CognitoIdentity identity,
             String functionVersion,
             String invokedFunctionArn,
-            ClientContext clientContext
+            ClientContext clientContext,
```
maxday, line 38: same here, maybe it would be better to swap args

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImplTest.java
```diff
@@ -333,6 +334,31 @@ public void nextTest() {
...
+    public void nextWithTenantIdTest() {
+        try {
+            MockResponse mockResponse = new MockResponse();
+            mockResponse.setResponseCode(HTTP_ACCEPTED);
+            mockResponse.setHeader("lambda-runtime-aws-request-id", "1234567890");
+            mockResponse.setHeader("Content-Type", "application/json");
+            String expectedTenantId = "my-tenant-id";
+            mockResponse.setHeader("lambda-runtime-aws-tenant-id", expectedTenantId);
```
maxday, line 345: I think it would be great to have another couple of tests for:
1. header not present
2. header present and empty string
3. header present and null value

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/include/aws/lambda-runtime/runtime.h
```diff
@@ -61,6 +61,11 @@ struct invocation_request {
      */
     std::chrono::time_point<std::chrono::system_clock> deadline;
 
+    /**
```
maxday, line 64: hopefully at some point, we won't have a copy of aws-lambda-cpp in here, so let's keep that change here but I think you also need to modify the source of truth here: https://github.com/awslabs/aws-lambda-cpp?

maxday (commented):
Awesome job @ramisa2108 ! 💯 
I left only some small comments, let me know if that makes sense

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/logging/JsonLogFormatterTest.java
```diff
@@ -29,7 +29,8 @@ void testFormattingWithLambdaContext() {
                 null,
                 null,
                 "function-arn",
-                null
+                null,
+                "tenant-id"
```
maxday, line 33: let's also add a test with a null value for tenant-id as it's optional

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/EventHandlerLoader.java
ramisa2108: Great suggestion. Updated.

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/api/LambdaContext.java
ramisa2108: Updated.

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImplTest.java
ramisa2108: Thanks for the suggestion. Added these test cases.

maxday (approved):
🎉

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/include/aws/lambda-runtime/runtime.h
godcrampy: Agreed, one day, we could get rid of this. (Not blocking PR from my end either).

godcrampy (approved):
Great work! 🎉

Issue #612, opened 2026-05-19
aws-lambda-java-log4j2 1.6.3 regression breaks logging


brandondahler:
### Problem
Log4j2 logging no longer works after upgrading to `com.amazonaws:aws-lambda-java-log4j2:1.6.3`.  

Log4j prints the error `ERROR Error processing element Lambda ([Appenders: null]): CLASS_NOT_FOUND` to stderr (which is not logged by lambda).

### Reproduction Steps
1. Create a new maven project as described below
2. Execute the code via `mvn package exec:java -Dexec.mainClass="bug.reproduction.App"`

#### Reproduction Files
`./pom.xml`
```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <groupId>bug.reproduction</groupId>
  <artifactId>aws-lambda-java-log4j2</artifactId>
  <version>1.0-SNAPSHOT</version>

  <name>aws-lambda-java-log4j2</name>

  <properties>
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <maven.compiler.release>17</maven.compiler.release>
  </properties>

  <dependencies>
    <dependency>
      <groupId>com.amazonaws</groupId>
      <artifactId>aws-lambda-java-log4j2</artifactId>
      <version>1.6.3</version> <!-- 1.6.2 works -->
    </dependency>
    <dependency>
      <groupId>org.apache.logging.log4j</groupId>
      <artifactId>log4j-core</artifactId>
      <version>2.25.4</version>
    </dependency>
    <dependency>
      <groupId>org.apache.logging.log4j</groupId>
      <artifactId>log4j-api</artifactId>
      <version>2.25.4</version>
    </dependency>
    <dependency>
      <groupId>org.apache.logging.log4j</groupId>
      <artifactId>log4j-layout-template-json</artifactId>
      <version>2.25.4</version>
    </dependency>
  </dependencies>

  <build>
    <pluginManagement>
      <plugins>
        <plugin>
          <artifactId>maven-clean-plugin</artifactId>
          <version>3.4.0</version>
        </plugin>
        <plugin>
[...]

PR #615, opened 2026-05-23
test(log4j2): Add initial test suite covering plugin registration

rschmitt:
This package previously had no tests and no JUnit dependency. Add JUnit Jupiter, the surefire plugin, and an initial test class that exercises the full plugin resolution path end-to-end: a log4j2.xml on the test classpath, a real LogManager-issued logger, and stdout capture (LambdaAppender writes through LambdaRuntime.getLogger() to System.out).

The TEXT test routes the root logger through LambdaTextFormat with a deterministic PatternLayout and asserts each level appears. The JSON test adds a second LambdaAppender with format="JSON" backed by JsonTemplateLayout + LambdaLayout.json, attached via additivity=false to a "json-test" logger, and asserts the messages show up JSON-encoded.

Add log4j-layout-template-json at test scope so the JSON path can resolve at test time. Users are still expected to bring their own copy at runtime (peer-dependency model, like PatternLayout); the published artifact's dependency surface is unchanged.

The tests succeed on Java 8 and fail on Java 25, because annotation processors are not run by default on Java 25, Log4j2Plugins.dat is not generated for our plugins, and both tests surface the resulting CLASS_NOT_FOUND from the log4j status logger.

### DIFF
--- a/aws-lambda-java-log4j2/pom.xml
+++ b/aws-lambda-java-log4j2/pom.xml
@@ -35,6 +35,7 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <log4j.version>2.25.4</log4j.version>
+        <junit-jupiter.version>5.12.2</junit-jupiter.version>
     </properties>
 
     <distributionManagement>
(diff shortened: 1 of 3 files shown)


darklight3it:
This should be merged before https://github.com/aws/aws-lambda-java-libs/pull/616

that is needed to prevent situations like https://github.com/aws/aws-lambda-java-libs/issues/612

darklight3it (commented):
LGTM 🚀  Thanks for the help!

M-Elsaeed (approved):

darklight3it (approved):

PR #617, opened 2026-05-23
Pin compilation to JDK8 via Maven Toolchains

darklight3it:
*Issue #, if available:*

*Description of changes:*

Add maven-toolchains-plugin to all packages requiring JDK [1.8,9). Builds now fail fast if no matching JDK 8 toolchain is configured, preventing silent issues like missing annotation processor output when building with higher JDKs.

*Target (OCI, Managed Runtime, both):* both

### DIFF
--- a/aws-lambda-java-events/pom.xml
+++ b/aws-lambda-java-events/pom.xml
@@ -42,6 +42,32 @@
         <junit-jupiter.version>5.12.2</junit-jupiter.version>
     </properties>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-toolchains-plugin</artifactId>
+                <version>3.2.0</version>
+                <configuration>
+                    <toolchains>
+                        <jdk>
+                            <!-- Range matches both "8" (e.g. actions/setup-java)
+                                 and "1.8" (legacy / hand-written toolchains). -->
+                            <version>[1.8,9)</version>
+                        </jdk>
+                    </toolchains>
+                </configuration>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>toolchain</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
     <distributionManagement>
         <repository>
             <id>sonatype-nexus-staging</id>
--- a/aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/Dockerfile.agent
+++ b/aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/Dockerfile.agent
@@ -15,4 +15,17 @@ RUN mkdir /apache-maven && \
 
 # Declare JDK 8 in toolchains.xml so maven-toolchains-plugin can resolve it
 RUN mkdir -p /root/.m2 && \
-    printf '<?xml version="1.0" encoding="UTF-8"?>\n<toolchains>\n  <toolchain>\n    <type>jdk</type>\n    <provides>\n      <version>8</version>\n    </provides>\n    <configuration>\n      <jdkHome>/usr/lib/jvm/java-1.8.0-amazon-corretto</jdkHome>\n    </configuration>\n  </toolchain>\n</toolchains>\n' > /root/.m2/toolchains.xml
+    cat > /root/.m2/toolchains.xml <<EOF
+<?xml version="1.0" encoding="UTF-8"?>
+<toolchains>
+  <toolchain>
+    <type>jdk</type>
+    <provides>
+      <version>8</version>
+    </provides>
+    <configuration>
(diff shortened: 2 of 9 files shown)


aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/Dockerfile.agent
```diff
@@ -11,4 +11,8 @@ COPY --from=docker/buildx-bin:latest /buildx /usr/libexec/docker/cli-plugins/doc
 ENV PATH="$PATH:/apache-maven/bin"
 RUN mkdir /apache-maven && \
     curl https://archive.apache.org/dist/maven/maven-3/3.8.7/binaries/apache-maven-3.8.7-bin.tar.gz | \
-    tar -xz -C /apache-maven --strip-components 1
\ No newline at end of file
+    tar -xz -C /apache-maven --strip-components 1
+
+# Declare JDK 8 in toolchains.xml so maven-toolchains-plugin can resolve it
+RUN mkdir -p /root/.m2 && \
+    printf '<?xml version="1.0" encoding="UTF-8"?>\n<toolchains>\n  <toolchain>\n    <type>jdk</type>\n    <provides>\n      <version>8</version>\n    </provides>\n    <configuration>\n      <jdkHome>/usr/lib/jvm/java-1.8.0-amazon-corretto</jdkHome>\n    </configuration>\n  </toolchain>\n</toolchains>\n' > /root/.m2/toolchains.xml
```
rudraroop: Two points:

1. Minor nit -> hardcoding the jdkHome path -> Unlikely, but possible that base image changes where it installs Corretto 8 (different version suffix or directory). Then this would break as the path won't resolve - is there a better way to ensure coupling?

2. Readability of the printf line -> You can use a *heredoc* (`<<`) instead of a one-line unformatted xml string. That way the xml would appear properly formatted

Functionally, everything looks okay though

aws-lambda-java-events/pom.xml
```diff
@@ -42,6 +42,32 @@
...
+                    <execution>
+                        <goals>
+                            <goal>toolchain</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
```
benrkia, line 70: Nit: i see this is repeated across all the pom files, is it worth considering having a shared parent pom and have all the common stuff in there?

rudraroop (commented):
Looks okay. Only one actionable feedback for readability

aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/Dockerfile.agent
benrkia: I'd create a file and do a copy instead of print if, that is better for maintainability

benrkia: Also let's add an example tollchains.xml (toolchains.xml.example) for contributors

rudraroop: @benrkia I suggested a heredoc for readability instead of having a separate file + copy as it's only 10 lines of xml. What do you think?

aws-lambda-java-events/pom.xml
fabisev: we have considered this. it might make sense when we automate the deployment, but for now each module is independently versioned and released to Maven Central and the CI builds them in isolation. a parent POM introduces coupling (any parent change needs to be published to maven central before any child module can be released). for this PR it feels safer to keep the modules self contained

aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/Dockerfile.agent
fabisev: > I'd create a file and do a copy instead of print if, that is better for maintainability

I believe that his Dockerfile is built via stdin (
```
.PHONY: setup-codebuild-agent
setup-codebuild-agent:
	docker build -t codebuild-agent \
	 --build-arg ARCHITECTURE=$(ARCHITECTURE_ALIAS) \
	  - < test/integration/codebuild-local/Dockerfile.agent
```
 in the Makefile), which means there is no build context to COPY for local files won't work. we would need to refactor the Makefile to use -f with a build context directory to enable that. 

right now i have switched to a heredoc with ${JAVA_HOME}, but happy to do the Makefile refactor if you think it would be nicer

benrkia: I'm fine with that but let's add an example file and mention that in contribution.md so that people don't get failure because toolchain definition is missing

rudraroop (approved):
LGTM

benrkia (approved):

PR #619, opened 2026-06-29
Add end-to-end integration tests

fabisev:
Add end-to-end integration tests for aws-lambda-java-log4j2
Adds integration tests that deploy a Lambda function using aws-lambda-java-log4j2, invoke it, and verify logs arrive in CloudWatch. This catches the main regression path: Log4j2 plugin discovery failing due to a missing Log4j2Plugins.dat in the shaded JAR. 

- lambda-integration-tests/ - SAM project with a minimal Java Lambda function that logs a unique marker at INFO/DEBUG/WARN/ERROR levels via Log4j2 LambdaAppender
- run-tests.sh - invokes the function and polls CloudWatch to verify the marker arrived
- build-integration-test.yml - PR gate that verifies the test function compiles
- run-integration-test.yml - deploys to AWS, runs the test, cleans up 

### DIFF
--- a/.github/workflows/build-integration-test.yml
+++ b/.github/workflows/build-integration-test.yml
@@ -26,10 +26,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Set up JDK
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@1bcf9fb12cf4aa7d266a90ae39939e61372fe520 # v5.4.0
         with:
           java-version: |
             8
--- a/.github/workflows/build-integration-test.yml
+++ b/.github/workflows/build-integration-test.yml
@@ -1,0 +1,56 @@
+# this workflow verifies that the integration test Lambda function builds successfully.
+# it does NOT deploy or run the tests (that requires AWS credentials and is done in
+# run-integration-test.yml).
+
+name: Build integration tests
+
+on:
+  push:
+    branches: [ main ]
+    paths:
+      - 'aws-lambda-java-log4j2/**'
+      - 'aws-lambda-java-core/**'
+      - 'lambda-integration-tests/**'
+  pull_request:
+    branches: [ '*' ]
+    paths:
+      - 'aws-lambda-java-log4j2/**'
+      - 'aws-lambda-java-core/**'
+      - 'lambda-integration-tests/**'
+      - '.github/workflows/build-integration-test.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up JDK
+        uses: actions/setup-java@v5
+        with:
+          java-version: |
+            8
+            21
+          distribution: corretto
+          cache: maven
+
+      - name: Install core with Maven
+        run: |
+          export JAVA_HOME=$JAVA_HOME_8_X64
+          mvn -B install --file aws-lambda-java-core/pom.xml
+
+      - name: Install log4j2 with Maven
+        run: |
+          export JAVA_HOME=$JAVA_HOME_8_X64
+          mvn -B install --file aws-lambda-java-log4j2/pom.xml
+
+      # build the integration test function
+      # this verifies that the function compiles and packages correctly.
+      # the tests will run in run-integration-test.yml which deploys to AWS.
+      - name: Package integration test function
+        run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/run-integration-test.yml
```diff
@@ -0,0 +1,102 @@
...
+
+permissions:
+  id-token: write
+  contents: read
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ main ]
+    paths:
```
darklight3it, line 14: Same here the test should be run for most packages, at least the one that are involved.

core, serialization, runtime-interface-client, log4j and i would also add events.

.github/workflows/build-integration-test.yml
```diff
@@ -0,0 +1,56 @@
+# this workflow verifies that the integration test Lambda function builds successfully.
+# it does NOT deploy or run the tests (that requires AWS credentials and is done in
+# run-integration-test.yml).
```
darklight3it, line 3: Aren't those supposed to fail as well in the run-integration-test. It seems to me that those test s are rebuilt also there.

.github/workflows/run-integration-test.yml
```diff
@@ -0,0 +1,102 @@
...
+    if: ${{ github.repository_owner == 'aws' }}
+    runs-on: ubuntu-latest
+    concurrency:
+      group: integration-test
+      cancel-in-progress: false
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up JDK
+        uses: actions/setup-java@v5
```
darklight3it: Same here and in all other references to github actions.

.github/workflows/run-integration-test.yml
```diff
@@ -0,0 +1,102 @@
...
+          aws-region: ${{ secrets.AWS_REGION }}
+
+      - name: Install core with Maven
+        run: |
+          export JAVA_HOME=$JAVA_HOME_8_X64
+          mvn -B install --file aws-lambda-java-core/pom.xml
+
+      - name: Install log4j2 with Maven
+        run: |
+          export JAVA_HOME=$JAVA_HOME_8_X64
```
darklight3it: It's probably correct here. But it is not correct in general.

In Java we have some C++ component that are dependent on the architecture. We should have a look at that together. 

The test should be run per arch and probably runtime (I was doing a CR also for that). For the moment it's probably fine.

We should also do this #617

.github/workflows/run-integration-test.yml
```diff
@@ -0,0 +1,102 @@
...
+jobs:
+  run-integration-tests:
+   # Only run on the main repo, not forks
+    if: ${{ github.repository_owner == 'aws' }}
+    runs-on: ubuntu-latest
+    concurrency:
+      group: integration-test
+      cancel-in-progress: false
+    steps:
+      - uses: actions/checkout@v6
```
darklight3it: add the commit sha to prevent supply chain attacks.

lambda-integration-tests/run-tests.sh
```diff
@@ -0,0 +1,103 @@
+# integration test script for aws-lambda-java-log4j2.
+# invokes the deployed lambda function and verifies logs appear in CloudWatch.
+
+set -euo pipefail
```
darklight3it, line 4: For the moment it's sufficient like this but we probably can do something more advanced. For example we can create a Java project that uses a something like JUNIT to run the AWS SDK invoke the function and then get the logs. 

In this way we avoid doing scripting and we can also have a nice report for how the test has gone.

.github/workflows/build-integration-test.yml
```diff
@@ -0,0 +1,56 @@
...
+          cache: maven
+
+      - name: Install core with Maven
+        run: |
+          export JAVA_HOME=$JAVA_HOME_8_X64
+          mvn -B install --file aws-lambda-java-core/pom.xml
+
+      - name: Install log4j2 with Maven
+        run: |
+          export JAVA_HOME=$JAVA_HOME_8_X64
```
maxday: let's make sure we run integ test on both architectures. You can use a matrix job to do this (ie: https://github.com/aws/aws-lambda-ruby-runtime-interface-client/blob/main/.github/workflows/integration-tests.yml#L32)

.github/workflows/build-integration-test.yml
darklight3it: This is solved. It's because we need to verify that the test are building, before merging that to main.

.github/workflows/run-integration-test.yml
darklight3it: Also we are going to deal with this later.

darklight3it (approved):
LGTM 🚀

anzheyazzz (approved):

maxday (approved):

PR #622, opened 2026-07-02
Fix log4j2 integration test deploy

fabisev:
*Description of changes:*
1. resolve_s3 conflict - 'samconfig.toml' had 'resolve_s3 = true' which conflicted with the '--s3-bucket' flag passed in the workflow.
2. explicit 'FunctionName' in the SAM template to avoid CloudFormation name truncation causing DELETE_FAILED on cleanup
3. stack status verification - after 'sam deploy' the workflow now checks that the stack is in 'CREATE_COMPLETE' or 'UPDATE_COMPLETE' before proceeding. If not, it logs the stack status and any failed resource events then exists 1. This prevents the cascading failure where a failed deploy leads to a confusing "env var is required" error in run-tests.sh. 


darklight3it (approved):
LGTM 🚀

maxday (approved):

PR #624, opened 2026-07-21
feat(runtime-interface-client): Add Lambda-Runtime-Invocation-Id support for cross-wiring protection

vip-amzn:
## Summary

Add `Lambda-Runtime-Invocation-Id` header support for cross-wiring protection.

The RIC now echoes the invocation ID received from RAPID on `/next` back on `/response` and `/error`, enabling RAPID to detect and reject stale responses from timed-out invocations.

## Problem

On Lambda Managed Instances (LMI) and On-Demand (OD), when an invoke times out, the runtime process continues running in the background. If a new invoke arrives with the same `requestId`, RAPID accepts it. The still-running old invocation eventually posts its response, and RAPID matches it to the new invoke — delivering the wrong response (cross-wiring).

## Solution

RAPID sends a unique per-invoke identifier via `Lambda-Runtime-Invocation-Id` header on `/next`. The runtime echoes it back on `/response` and `/error`. RAPID validates the match before accepting the response.

## Changes

- `aws-lambda-cpp-0.2.7/runtime.h`: Added `invocation_id` field to `invocation_request`, added param to `post_success`/`post_failure`/`do_post`
- `aws-lambda-cpp-0.2.7/runtime.cpp`: Parse header in `get_next()`, add to curl request in `do_post()` when non-empty
- `NativeClient.cpp`/`.h`: JNI reads `invocation_id` from C++ response, passes to Java; `postInvocationResponse` accepts nullable invocationId
- `InvocationRequest.java`: New `invocationId` field
- `NativeClient.java`: Updated native method signature
- `LambdaRuntimeApiClient.java`/`Impl`: `reportInvocationSuccess`/`reportInvocationError` accept invocationId
- `AWSLambda.java`: Thread invocationId through runtime loop
- Unit tests: Updated assertions for new parameter
- Version bump: 2.11.0 → 2.12.0

## Backward Compatibility

Fully backward compatible in both directions:
- If RAPID doesn't send the header → RIC doesn't see it → doesn't echo → no behavior change
- If RIC doesn't echo it (old version) → RAPID skips validation → no behavior change

### DIFF
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImpl.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImpl.java
@@ -34,6 +34,7 @@ public class LambdaRuntimeApiClientImpl implements LambdaRuntimeApiClient {
     private static final String DEFAULT_CONTENT_TYPE = "application/json";
     private static final String XRAY_ERROR_CAUSE_HEADER = "Lambda-Runtime-Function-XRay-Error-Cause";
     private static final String ERROR_TYPE_HEADER = "Lambda-Runtime-Function-Error-Type";
+    private static final String INVOCATION_ID_HEADER = "Lambda-Runtime-Invocation-Id";
     // 1MiB
     private static final int XRAY_ERROR_CAUSE_MAX_HEADER_SIZE = 1024 * 1024;
     
@@ -55,7 +56,7 @@ public class LambdaRuntimeApiClientImpl implements LambdaRuntimeApiClient {
     @Override
     public void reportInitError(LambdaError error) throws IOException {
         String endpoint = this.baseUrl + "/2018-06-01/runtime/init/error";
-        reportLambdaError(endpoint, error, XRAY_ERROR_CAUSE_MAX_HEADER_SIZE);
+        reportLambdaError(endpoint, error, XRAY_ERROR_CAUSE_MAX_HEADER_SIZE, null);
     }
 
     @Override
@@ -123,14 +124,15 @@ public class LambdaRuntimeApiClientImpl implements LambdaRuntimeApiClient {
     }
 
     @Override
-    public void reportInvocationSuccess(String requestId, byte[] response) {
-        NativeClient.postInvocationResponse(requestId.getBytes(UTF_8), response);
+    public void reportInvocationSuccess(String requestId, byte[] response, String invocationId) {
+        byte[] invocationIdBytes = invocationId != null ? invocationId.getBytes(UTF_8) : null;
+        NativeClient.postInvocationResponse(requestId.getBytes(UTF_8), response, invocationIdBytes);
     }
 
     @Override
-    public void reportInvocationError(String requestId, LambdaError error) throws IOException {
+    public void reportInvocationError(String requestId, LambdaError error, String invocationId) throws IOException {
         String endpoint = invocationEndpoint + requestId + "/error";
-        reportLambdaError(endpoint, error, XRAY_ERROR_CAUSE_MAX_HEADER_SIZE);
+        reportLambdaError(endpoint, error, XRAY_ERROR_CAUSE_MAX_HEADER_SIZE, invocationId);
     }
 
     @Override
@@ -145,12 +147,16 @@ public class LambdaRuntimeApiClientImpl implements LambdaRuntimeApiClient {
     @Override
     public void reportRestoreError(LambdaError error) throws IOException {
         String endpoint = this.baseUrl + "/2018-06-01/runtime/restore/error";
-        reportLambdaError(endpoint, error, XRAY_ERROR_CAUSE_MAX_HEADER_SIZE);
+        reportLambdaError(endpoint, error, XRAY_ERROR_CAUSE_MAX_HEADER_SIZE, null);
     }
 
-    void reportLambdaError(String endpoint, LambdaError error, int maxXrayHeaderSize) throws IOException {
+    void reportLambdaError(String endpoint, LambdaError error, int maxXrayHeaderSize, String invocationId) throws IOException {
         Map<String, String> headers = new HashMap<>();
         headers.put(ERROR_TYPE_HEADER, error.errorType.getRapidError());
+
+        if (invocationId != null) {
+            headers.put(INVOCATION_ID_HEADER, invocationId);
+        }
 
         if (error.xRayErrorCause != null) {
             byte[] xRayErrorCauseJson = DtoSerializers.serialize(error.xRayErrorCause);
--- a/aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
+++ b/aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
@@ -41,6 +41,7 @@ static constexpr auto COGNITO_IDENTITY_HEADER = "lambda-runtime-cognito-identity
 static constexpr auto DEADLINE_MS_HEADER = "lambda-runtime-deadline-ms";
 static constexpr auto FUNCTION_ARN_HEADER = "lambda-runtime-invoked-function-arn";
 static constexpr auto TENANT_ID_HEADER = "lambda-runtime-aws-tenant-id";
+static constexpr auto INVOCATION_ID_HEADER = "lambda-runtime-invocation-id";
 thread_local static CURL* m_curl_handle = curl_easy_init();
 
 enum Endpoints {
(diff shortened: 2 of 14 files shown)


aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImpl.java
```diff
@@ -145,13 +146,17 @@ public void restoreNext() throws IOException {
...
+        reportLambdaError(endpoint, error, XRAY_ERROR_CAUSE_MAX_HEADER_SIZE, null);
     }
 
-    void reportLambdaError(String endpoint, LambdaError error, int maxXrayHeaderSize) throws IOException {
+    void reportLambdaError(String endpoint, LambdaError error, int maxXrayHeaderSize, String invocationId) throws IOException {
         Map<String, String> headers = new HashMap<>();
         headers.put(ERROR_TYPE_HEADER, error.errorType.getRapidError());
 
+        if (invocationId != null) {
+            headers.put("Lambda-Runtime-Invocation-Id", invocationId);
```
rudraroop: the other headers in this file are declared as constants at the start of the file. I would suggest keeping that uniform and making a constant above for this one as well

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
```diff
@@ -41,6 +41,7 @@ static constexpr auto COGNITO_IDENTITY_HEADER = "lambda-runtime-cognito-identity
 static constexpr auto DEADLINE_MS_HEADER = "lambda-runtime-deadline-ms";
 static constexpr auto FUNCTION_ARN_HEADER = "lambda-runtime-invoked-function-arn";
 static constexpr auto TENANT_ID_HEADER = "lambda-runtime-aws-tenant-id";
+static constexpr auto INVOCATION_ID_HEADER = "lambda-runtime-invocation-id";
```
rudraroop, line 44: The "L" is capitalised in the java file but lowercase here

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImplTest.java
```diff
@@ -274,7 +274,7 @@ public void reportInvocationErrorTestWrongStatusCode() {
             mockWebServer.enqueue(mockResponse);
 
             LambdaError lambdaError = new LambdaError(errorRequest, rapidErrorType);
-            lambdaRuntimeApiClientImpl.reportInvocationError(requestId, lambdaError);
+            lambdaRuntimeApiClientImpl.reportInvocationError(requestId, lambdaError, null);
```
rudraroop, line 277: I think we should have one test for `reportInvocationError` which tests a positive path i.e. one where the `Lambda-Runtime-Invocation-Id` is actually getting passed instead of null

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambdaTest.java
rudraroop, line 1: Same feedback for positive path test inclusion as in this file [LambdaRuntimeApiClientImplTest.java‎](https://github.com/aws/aws-lambda-java-libs/pull/624/changes#diff-d77df1af22dec4b64021992568626262be2057b043966f7ad36793ffd4146dcd)

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
```diff
@@ -341,6 +347,9 @@ runtime::post_outcome runtime::do_post(
     headers = curl_slist_append(headers, "Expect:");
     headers = curl_slist_append(headers, "transfer-encoding:");
     headers = curl_slist_append(headers, m_user_agent_header.c_str());
+    if (!invocation_id.empty()) {
+        headers = curl_slist_append(headers, ("lambda-runtime-invocation-id: " + invocation_id).c_str());
```
rudraroop: Should use the header constant here instead of hardcoding again?

rudraroop (commented):
Looks good overall - my major point of feedback would be adding a test for the positive path where invocation IDs are non-null

Have added some casing/hardcoding related comments - whatever I could see

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImpl.java
darklight3it: +1

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
darklight3it: +1 but I'm not sure if header module in java applies a lowercase transform

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
```diff
@@ -41,6 +41,7 @@ static constexpr auto COGNITO_IDENTITY_HEADER = "lambda-runtime-cognito-identity
 static constexpr auto DEADLINE_MS_HEADER = "lambda-runtime-deadline-ms";
```
darklight3it, line 41: general question we need to investigate. Why does the jni module source is in a version tagged folder (`aws-lambda-cpp-0.2.7`), do we expect this to change in the current release model? Is this intentional?

This can also be interesting for @fabisev

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
darklight3it: +1

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImplTest.java
darklight3it: +1

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambdaTest.java
darklight3it: +1

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
vip-amzn: HTTP headers are case-insensitive. The Java file uses title-case to match the convention of the other constants in that file. while the C++ uses lowercase because that's the convention in aws-lambda-cpp.

aws-lambda-java-runtime-interface-client/src/main/jni/deps/aws-lambda-cpp-0.2.7/src/runtime.cpp
vip-amzn: Done

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/LambdaRuntimeApiClientImplTest.java
vip-amzn: added `reportInvocationErrorWithInvocationIdTest` in LambdaRuntimeApiClientImplTest.java.

aws-lambda-java-runtime-interface-client/src/test/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambdaTest.java
vip-amzn: Added testInvocationIdIsPassedToReportError and testInvocationIdIsPassedToReportSuccess tests.

rudraroop (approved):
Looks good. Thanks for the changes

aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/Dockerfile.agent
```diff
@@ -1,9 +1,7 @@
 FROM public.ecr.aws/amazoncorretto/amazoncorretto:8
 
 # Install docker and buildx extension
-RUN amazon-linux-extras enable docker && \
```
vip-amzn, line 4: Failing smoke test on Java RIC PR:

### What happened:
The Dockerfile.agent at test/integration/codebuild-local/Dockerfile.agent (on main: https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-runtime-interface-client/test/integration/codebuild-local/Dockerfile.agent#L4) uses amazon-linux-extras enable docker which doesn't exist on AL2023.

### Proof — same Dockerfile, different base OS:
• [Successful job](https://github.com/aws/aws-lambda-java-libs/actions/runs/29409951896/job/87334403290#step:6:3314) (before Jul 21): sha256:cbe7fe... → Amazon Linux 2
• [Failed job](https://github.com/aws/aws-lambda-java-libs/actions/runs/30082779342/job/89447952463?pr=624#step:6:3303) (after Jul 21): sha256:f4ee79... → Amazon Linux 2023

### Corretto announcement:
https://github.com/corretto/corretto-docker#notice-default-images-moved-from-amazon-linux-2-to-amazon-linux-2023
> Amazon Linux 2 (AL2) is now end of life. The default amazoncorretto images (for example amazoncorretto:8, amazoncorretto:11, amazoncorretto:17, amazoncorretto:21 and latest) are now based on Amazon Linux 2023 (AL2023). Corretto 22+ images were already AL2023 based.

### Who uses it:
Only make test-smoke / make test-integ (local CodeBuild agent for integration tests). The actual make test and make build targets that compile and run unit tests pass fine — those don't use this Dockerfile.

### Fix options:
1. Change base to amazoncorretto:8-al2 (fallback tag, temporary)
2. Remove amazon-linux-extras enable docker line (docker is in default repos on AL2023)

Going ahead with Option 2.

On AL2023, `xargs` is no longer included by default (it was part of `findutils` on AL2 which was pre-installed). The test script uses xargs but the codebuild agent image doesn't have it.

darklight3it (approved):

PR #626, opened 2026-07-24
Fabisev/automating mvn deployment

fabisev:
*Issue #, if available:*

*Description of changes:*

Replaces the manual `mvn deploy` release with GitHub Actions release pipelines that build, test, sign, and publish CI-validated artifacts to Maven Central. 

- `release.yml`: `workflow_dispatch` pipeline for the six pure-Java modules. Pinned JDK 8, `mvn verify` gate, then `release:prepare`/`release:perform` with a publish-safe ordering (prepare locally, publish, then `git push --atomic` only after publish succeeds; failure rolls back the runner and never modifies the remote). Repo-wide `release` concurrency group so releases never overlap.
- `release-runtime-interface-client.yml`: dedicated RIC pipeline. RIC ships a JNI native library for four targets plus a main JAR, so each native library is built on a runner of its own architecture (x86_64 on `ubuntu-latest`, aarch_64 on `ubuntu-24.04-arm`) instead of QEMU emulation, then one job assembles all artifacts and publishes them.
- POM changes across all modules: SNAPSHOT versions, `maven-release-plugin` configuration, SCM `developerConnection`, and `autoPublish=true`.
- Credentials are fetched at runtime via GitHub OIDC 

*Testing done:*

Validated locally with `act` (GitHub Actions run in a container)
- Workflow mechanics (dry-run of `release.yml`): checkout, pinned JDK 8, `mvn verify`, and `release:prepare -DdryRun` pass; correct release tag and next-development version bump.
- Publish chain (against a personal Sonatype namespace I control, with auto-publish disabled and the deployment dropped afterward): GPG signing, token auth, upload, and Sonatype validation all succeed with no errors, exercising the same signing/settings/publishing mechanic the real workflow uses.
- log4j2 module correctness (built on pinned JDK 8): the jar contains the `Log4j2Plugins.dat` descriptor, classes are Java 8 bytecode, and the descriptor unit test passes.
- RIC native build: the arm64 classifier jar was built on JDK 8 and contains a correct ARM aarch64 ELF shared object.
- RIC multi-artifact publish shape: a main jar plus four classifier jars deploy as a single deployment and validate with no errors.
- Credential fetch (read-only)

Not exercisable outside a real run in this repo: the live token exchange and the CI build-matrix assembly. These will be covered by the first dry-run (`skip_publish: true`) after merge.

*Target (OCI, Managed Runtime, both):* N/A

### DIFF
--- a/.github/workflows/release-runtime-interface-client.yml
+++ b/.github/workflows/release-runtime-interface-client.yml
@@ -69,12 +69,22 @@ jobs:
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
-      - name: Set up JDK 8
-        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5
-        with:
-          java-version: 8
-          distribution: corretto
-          cache: maven
+      # Use the CodeBuild image's preinstalled Corretto 8. The image ships it at
+      # $JAVA_8_HOME but defaults JAVA_HOME to Java 25, so point JAVA_HOME/PATH at
+      # 8. Avoids actions/setup-java, which fetches from corretto.github.io +
+      # corretto.aws, both blocked by the runner egress lock. $JAVA_8_HOME
+      # resolves per-arch (x86_64/aarch64).
+      - name: Use the runner image's preinstalled Corretto 8
+        run: |
+          echo "JAVA_HOME=$JAVA_8_HOME" >> "$GITHUB_ENV"
+          echo "$JAVA_8_HOME/bin" >> "$GITHUB_PATH"
+          "$JAVA_8_HOME/bin/java" -version
+
+      # Route all mvn resolution through the CodeArtifact mirror. Must precede
+      # resolve-release-version, which invokes `mvn help:evaluate`. Ambient
+      # CodeBuild runner-role creds supply the token; no OIDC step in this job.
+      - name: Configure Maven CodeArtifact mirror
+        uses: ./.github/actions/configure-maven-mirror
 
       - name: Resolve and validate release version
         uses: ./.github/actions/resolve-release-version
@@ -112,6 +122,12 @@ jobs:
             ${{ env.MODULE }}/target/*-linux*.jar
             ${{ env.MODULE }}/target/classes/jni/*.so
 
+      # Remove the user settings holding the CodeArtifact mirror token once the
+      # build is done. Ephemeral runner, so defence-in-depth, not load-bearing.
+      - name: Scrub Maven settings
+        if: always()
+        run: rm -f "$HOME/.m2/settings.xml"
+
   # Assemble all native builds and publish.
   release:
     needs: build-natives
@@ -123,12 +139,23 @@ jobs:
         with:
           fetch-depth: 0   # full history for tagging/pushing
 
-      - name: Set up JDK 8
-        uses: actions/setup-java@be666c2fcd27ec809703dec50e508c2fdc7f6654 # v5
-        with:
-          java-version: 8
-          distribution: corretto
-          cache: maven
+      # Use the CodeBuild image's preinstalled Corretto 8. The image ships it at
+      # $JAVA_8_HOME but defaults JAVA_HOME to Java 25, so point JAVA_HOME/PATH at
+      # 8. Avoids actions/setup-java, which fetches from corretto.github.io +
+      # corretto.aws, both blocked by the runner egress lock. $JAVA_8_HOME
+      # resolves per-arch (x86_64/aarch64).
+      - name: Use the runner image's preinstalled Corretto 8
+        run: |
+          echo "JAVA_HOME=$JAVA_8_HOME" >> "$GITHUB_ENV"
+          echo "$JAVA_8_HOME/bin" >> "$GITHUB_PATH"
+          "$JAVA_8_HOME/bin/java" -version
+
+      # Route all mvn resolution through the CodeArtifact mirror. Must precede
+      # resolve-release-version (which invokes `mvn help:evaluate`) and the OIDC
+      # step (which would shadow the runner-role creds this needs). Runs on every
+      # path, since dependency resolution happens on dry-runs too.
+      - name: Configure Maven CodeArtifact mirror
+        uses: ./.github/actions/configure-maven-mirror
 
       - name: Resolve and validate release version
         uses: ./.github/actions/resolve-release-version
@@ -234,7 +261,11 @@ jobs:
           gpg --batch --import <<< "$GPG_PRIVATE_KEY"
           GPG_KEYNAME=$(gpg --list-secret-keys --with-colons | awk -F: '/^sec:/ {print $5; exit}')
 
-          # settings.xml with the Sonatype token (server id "central").
+          # Global settings holding only the Sonatype "central" server for upload.
+          # Passed to Maven as -gs (global) so it MERGES with the CodeArtifact
+          # mirror in ~/.m2/settings.xml (user) that the mirror step wrote: deps
+          # resolve through the mirror, upload goes to central, and the mirror
+          # token stays in that user file instead of being re-passed here.
           {
             echo '<settings><servers><server>'
             echo "<id>central</id>"
@@ -243,9 +274,9 @@ jobs:
             echo '</server></servers></settings>'
           } > "$MAVEN_SETTINGS"
 
-          # --- Publish ---
+          # --- Publish --- (-gs: merge Sonatype creds with the ~/.m2 mirror)
           mvn deploy -Prelease -DskipTests -DmultiArch=false \
-            -s "$MAVEN_SETTINGS" \
+            -gs "$MAVEN_SETTINGS" \
             -Dgpg.keyname="$GPG_KEYNAME" -Dgpg.passphrase="$GPG_PASSPHRASE" \
             --file "$MODULE/pom.xml"
 
--- a/.github/workflows/release-runtime-interface-client.yml
+++ b/.github/workflows/release-runtime-interface-client.yml
@@ -49,10 +49,10 @@ jobs:
       matrix:
         include:
           - arch: x86_64
-            runner: ubuntu-latest
+            runner: codebuild-aws-lambda-java-libs-test-trigger-x86-${{ github.run_id }}-${{ github.run_attempt }}
             profiles: linux-x86_64 linux_musl-x86_64
           - arch: aarch64
-            runner: ubuntu-24.04-arm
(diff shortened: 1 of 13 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,235 @@
...
+      - name: Configure AWS credentials (OIDC)
+        if: ${{ github.event.inputs.skip_publish != 'true' }}
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4
+        with:
+          aws-region: ${{ env.AWS_REGION }}
+          role-to-assume: ${{ env.OIDC_ROLE_ARN }}
+          role-session-name: GitHubActionsMavenCentralRelease
+          role-duration-seconds: 3600
+
+      - name: Fetch signing key and Sonatype credentials
```
darklight3it: I think here we should collapse this step with the following. In this way we are not passing sensitive information across different boundaries using `$GITHUB_ENV`. A malevolent actor can leverage a step with lower permission to get that.

.github/workflows/release.yml
```diff
@@ -0,0 +1,235 @@
...
+          # settings.xml with the Sonatype token (server id "central").
+          SETTINGS="$RUNNER_TEMP/settings.xml"
+          {
+            echo '<settings><servers><server>'
+            echo "<id>central</id>"
+            echo "<username>${SONATYPE_USERNAME}</username>"
+            echo "<password>${SONATYPE_PASSWORD}</password>"
+            echo '</server></servers></settings>'
+          } > "$SETTINGS"
+
```
darklight3it: We should remove this file just after it is used in this way we are not leaving into the filesystem a file with sensitive informations in it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,235 @@
...
+          # Import the key with loopback pinentry so Maven can sign non-interactively.
+          GNUPGHOME=$(mktemp -d)
+          chmod 700 "$GNUPGHOME"
+          echo "allow-loopback-pinentry" > "$GNUPGHOME/gpg-agent.conf"
+          echo "pinentry-mode loopback" > "$GNUPGHOME/gpg.conf"
+          export GNUPGHOME
+          gpgconf --kill gpg-agent || true
+          gpg --batch --import <<< "$GPG_PRIVATE_KEY"
+          GPG_KEYNAME=$(gpg --list-secret-keys --with-colons | awk -F: '/^sec:/ {print $5; exit}')
```
darklight3it, line 211: Here we are blindly trusting whatever is coming from the secrets manager. We can make this code stronger catching someone that is trying to tamper the private keys in this way.

```sh
 gpg --batch --import <<< "$GPG_PRIVATE_KEY"

# Verify the imported key matches the expected fingerprint.
ACTUAL_FPR=$(gpg --list-secret-keys --with-colons | awk -F: '/^fpr:/ {print $10; exit}')
if [[ "$ACTUAL_FPR" != "$EXPECTED_GPG_FINGERPRINT" ]]; then
    echo "::error::GPG key fingerprint mismatch: expected $EXPECTED_GPG_FINGERPRINT, got $ACTUAL_FPR"
    exit 1
fi

GPG_KEYNAME=$(gpg --list-secret-keys --with-colons | awk -F: '/^sec:/ {print $5; exit}')
```

Essentially we double check that the gpg is conformant to what is published on the keyserver in its public part.

This is a little bit convoluted for now but we should consider doing that in a follow up.

.github/workflows/release.yml
```diff
@@ -0,0 +1,235 @@
...
+  AWS_REGION: ${{ vars.AWS_REGION_MAVEN_RELEASE }}
+  OIDC_ROLE_ARN: ${{ secrets.AWS_ROLE_MAVEN_RELEASE }}
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    environment: Release
+    timeout-minutes: 30
+
+    steps:
```
darklight3it, line 74: Since this is manual can we add a guard to avoid branch deploy?

```
- name: Verify release branch
    run: |
        if [[ "$GITHUB_REF_NAME" != "main" ]]; then
            echo "::error::Releases must run from the main branch, got '$GITHUB_REF_NAME'"
            exit 1
        fi
    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
        ...
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,235 @@
...
+        run: mvn verify --file "$MODULE/pom.xml"
+
+      - name: Configure AWS credentials (OIDC)
+        if: ${{ github.event.inputs.skip_publish != 'true' }}
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4
+        with:
+          aws-region: ${{ env.AWS_REGION }}
+          role-to-assume: ${{ env.OIDC_ROLE_ARN }}
+          role-session-name: GitHubActionsMavenCentralRelease
+          role-duration-seconds: 3600
```
darklight3it: 1 hour for this is definitely too long I think we can cut it to something like 300s (~5 min).

.github/workflows/release-runtime-interface-client.yml
```diff
@@ -0,0 +1,288 @@
...
+      - name: Configure AWS credentials (OIDC)
+        if: ${{ github.event.inputs.skip_publish != 'true' }}
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4
+        with:
+          aws-region: ${{ env.AWS_REGION }}
+          role-to-assume: ${{ env.OIDC_ROLE_ARN }}
+          role-session-name: GitHubActionsRicMavenCentralRelease
+          role-duration-seconds: 3600
+
+      - name: Fetch signing key and Sonatype credentials
```
darklight3it: I think there is a lot of space here to create a custom action to make this logic reusable between the two workflows.

.github/workflows/release-runtime-interface-client.yml
```diff
@@ -0,0 +1,288 @@
...
+            echo "<id>central</id>"
+            echo "<username>${SONATYPE_USERNAME}</username>"
+            echo "<password>${SONATYPE_PASSWORD}</password>"
+            echo '</server></servers></settings>'
+          } > "$SETTINGS"
+
+          echo "GNUPGHOME=$GNUPGHOME" >> "$GITHUB_ENV"
+          echo "GPG_KEYNAME=$GPG_KEYNAME" >> "$GITHUB_ENV"
+          echo "GPG_PASSPHRASE=$GPG_PASSPHRASE" >> "$GITHUB_ENV"
+          echo "MAVEN_SETTINGS=$SETTINGS" >> "$GITHUB_ENV"
```
darklight3it: Let's collaps also here.

.github/workflows/release-runtime-interface-client.yml
```diff
@@ -0,0 +1,288 @@
...
+
+env:
+  MODULE: aws-lambda-java-runtime-interface-client
+  RELEASE_VERSION_INPUT: ${{ github.event.inputs.releaseVersion }}
+  DEVELOPMENT_VERSION_INPUT: ${{ github.event.inputs.developmentVersion }}
+  MAVEN_ARGS: "-B --no-transfer-progress"
+  AWS_REGION: ${{ vars.AWS_REGION_MAVEN_RELEASE }}
+  OIDC_ROLE_ARN: ${{ secrets.AWS_ROLE_MAVEN_RELEASE }}
+
+jobs:
```
darklight3it, line 44: Add the guard here too.

.github/workflows/release-runtime-interface-client.yml
```diff
@@ -0,0 +1,288 @@
...
+          ls -1 "$MODULE/target/"*-linux*.jar "$MODULE/target/classes/jni/"*.so
+
+      - name: Configure AWS credentials (OIDC)
+        if: ${{ github.event.inputs.skip_publish != 'true' }}
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4
+        with:
+          aws-region: ${{ env.AWS_REGION }}
+          role-to-assume: ${{ env.OIDC_ROLE_ARN }}
+          role-session-name: GitHubActionsRicMavenCentralRelease
+          role-duration-seconds: 3600
```
darklight3it: Cut also this role duration.

.github/workflows/release.yml
```diff
@@ -0,0 +1,263 @@
...
+      # and Sonatype token stay in this shell and never cross a $GITHUB_ENV
+      # boundary, where a later (possibly compromised) step could read them.
+      # prepare/perform aren't atomic: prepare locally, publish, push only after.
+      - name: Release (prepare locally, publish, then push)
+        if: ${{ github.event.inputs.skip_publish != 'true' }}
+        run: |
+          # Scrub the settings.xml (contains the Sonatype token) and the keyring
+          # on exit, so no sensitive file is left on the runner even on failure.
+          MAVEN_SETTINGS="$RUNNER_TEMP/settings.xml"
+          export GNUPGHOME=$(mktemp -d)
```
darklight3it, line 191: Really good idea using an ephemeral folder 🚀 , this gives the guarantee of isolation on single jobs. We can probably use the same for `$MAVEN_SETTINGS`, since `$RUNNERT_TEMP` is phisical and we don't need this to survive that long.

.github/workflows/release.yml
```diff
@@ -0,0 +1,263 @@
...
+          trap 'rm -rf "$MAVEN_SETTINGS" "$GNUPGHOME"' EXIT
```
darklight3it, line 192: Also good idea using the trap here 🚀 . This cleans the data if there is a failure along the line.

darklight3it (approved):
LGTM 🚀

rudraroop (commented):
Overall, looks good - I do have a doubt about the _reusability_ of this workflow

If you look at this file - [run-integration-test.yml](https://github.com/aws/aws-lambda-java-libs/blob/feat/invocation-id-cross-wiring/.github/workflows/run-integration-test.yml#L10) 

It contains this block
```
on:
  workflow_dispatch:
....
```

From what I understand this allows workflows to be used but not reused or called by other workflows. I believe an additional block that looks like

```
on:
  workflow_call:
....
```
May be required. This is the impression I got from the official documentation - https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows - but I'm not sure if this repository handles things differently somehow

In any case, worth checking and confirming before we merge

.github/workflows/run-integration-test.yml
```diff
@@ -9,6 +9,7 @@ permissions:
 
 on:
   workflow_dispatch:
+  workflow_call:
```
rudraroop, line 12: Neat

rudraroop (approved):
LGTM

PR #629, opened 2026-08-13
fix: bypass proxy for Runtime API HTTP calls

benrkia:
*Issue #, if available:*

*Description of changes:*

Use Proxy.NO_PROXY when opening HttpURLConnection to RAPID endpoints. This prevents customer-configured http.proxyHost from routing internal Runtime API calls through the proxy, which fails when RAPID is not on a loopback address (localhost is implicitly exempt in most proxy implementations).

The hot path (nextInvocation/postInvocationResponse) already bypasses Java's ProxySelector via the native C++/libcurl NativeClient and is not affected.

Proxy.NO_PROXY is per-connection scoped, zero impact on customer handler HTTP calls.

*Target : both OCI and Managed Runtime*


darklight3it:
Code is good 👍 .

Please update the release notes with the fix and update your branch.

darklight3it (approved):

maxday (commented):
LGTM!

maxday (approved):

PR #634, opened 2026-08-18
fix: speed up CI

maxday:
Speeds up the Runtime Interface Client (RIC) CI by running each architecture on a native GitHub runner instead of cross-building both architectures on a single x86_64 runner under QEMU emulation.

Previously the `build` and `smoke-test` jobs ran on `ubuntu-latest` and used QEMU to emulate `linux/arm64/v8`. Under emulation `mvn install` recompiles curl for aarch64, which took ~30 minutes. This PR splits the work per-architecture so the aarch64 half runs on a native ARM runner (`ubuntu-24.04-arm`), eliminating the emulation cost.

*Results:*

Measured on the RIC PR workflow (both jobs run in parallel, so the reported wall-clock is the longest job).

| Job | Before (QEMU, single x86_64 runner) | After (native runners) |
| --- | --- | --- |
| `build` | 44m 06s | x86_64: 5m 31s · aarch64: 4m 47s |
| `smoke-test` | 45m 51s | x86_64: 8m 55s · aarch64: 9m 24s |
| **Workflow wall-clock** | **~46 min** | **~13 min** |

Net gain: **~33 minutes faster (~71% reduction)** in end-to-end CI time. The bulk of the saving comes from `smoke-test`, which dropped from ~46 min to ~9 min (~37 min, ~80% faster) by eliminating the emulated aarch64 curl recompile.

### DIFF
--- a/.github/workflows/runtime-interface-client_pr.yml
+++ b/.github/workflows/runtime-interface-client_pr.yml
@@ -46,7 +46,15 @@ jobs:
         IS_JAVA_8: true
 
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-latest
+          - arch: aarch64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
     steps:
     - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
@@ -57,16 +65,10 @@ jobs:
         distribution: corretto
         cache: maven
 
-    - name: Set up QEMU
-      uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3
-
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
       with:
         install: true
-
-    - name: Available buildx platforms
-      run: echo ${{ steps.buildx.outputs.platforms }}
 
     - name: Build and install core dependency locally
       working-directory: ./aws-lambda-java-core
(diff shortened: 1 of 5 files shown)


vip-amzn (approved):

fabisev:
Nice speedup! 

One thing I noticed is that making build and smoke-test matrix jobs changes the check names to build (x86_64) etc. Main requires a check named just build, so it's stuck pending. 
One thing I did for the build-integration-test.yml workflow is:

```
build:
    needs: build-arch
    if: always()
    runs-on: ubuntu-latest
    steps:
      - name: Check build results
        run: |
          if [ "${{ needs.build-arch.result }}" != "success" ]; then
            echo "Build failed on one or more architectures"
            exit 1
          fi
          
```
So a build-arch job with a build job aggregator, that would keep the build check reporting without touching the branch protection.

fabisev (approved):

fabisev (approved):

darklight3it (commented):
Nice but I think we need to check something more here for example this. I think we can probably get rid of it and all the dockeremu pipeline right?

https://github.com/aws/aws-lambda-java-libs/blob/6cdbbdc6ed207bc788872a8d0e11794d97d6f3a2/aws-lambda-java-runtime-interface-client/pom.xml#L46

maxday:
@darklight3it I think your comment got truncated?
I agree there's more cleanup we could do, but this PR delivers a huge performance win on its own, and I'd love to get it merged ASAP so we can start benefiting from it. We can then handle the remaining QEMU config in follow-up PRs. WDYT? I've tackled par of the removal in https://github.com/aws/aws-lambda-java-libs/pull/633

vip-amzn (approved):

darklight3it:
@maxday fine for me as long as we create the new PR immediately I don't want dead code to stay long. The package is already huge and confusing.

PR #635, opened 2026-08-19
Jackson version bump to 2.18.9

rudraroop:
*Issue #, if available:*

*Description of changes:*
Upgrade `jackson` dependency version to `2.18.9` in the serialization package to address the following CVE reports [CVE-2026-54512](https://access.redhat.com/security/cve/cve-2026-54512) and [CVE-2026-54513](https://access.redhat.com/security/cve/cve-2026-54513)

Will update tests package separately after maven workflow's auto version bump for serialization


aws-lambda-java-tests/pom.xml
```diff
@@ -43,7 +43,7 @@
         -->
         <junit.version>5.9.2</junit.version>
         <jacoco.maven.plugin.version>0.8.7</jacoco.maven.plugin.version>
-        <aws-lambda-java-serialization.version>1.4.1</aws-lambda-java-serialization.version>
```
darklight3it, line 46: It's fine but we need to remember of releasing also this package in the followup PR.

darklight3it (approved):

aws-lambda-java-tests/pom.xml
darklight3it: actually I was wrong. The test are exercised locally but you need to have the same version, otherwise it will fail the test, because it is registered locally as 1.4.1.

rudraroop: Cool, I'll revert this then

maxday (approved):

PR #640, opened 2026-08-24
feat(release): add changelog entry, GitHub Releases, lastPublished, a…

fabisev:
*Issue #, if available:* N/A

*Description of changes:*

Adds changelog and release-visibility improvements to both Maven Central release workflows, pulls the shared release logic into composite actions, and tightens up version and secret handling. No change to the build itself.

**`.github/workflows/release.yml`** (pure-Java modules)

- New required `changelogEntry` input. It's prepended to the module's `RELEASE.CHANGELOG.md` (committed into the post-release version-bump PR) and reused as the GitHub Release notes. Passed via env, never interpolated into the shell.
- New step to create a **GitHub Release** on the pushed tag. It includes:
  - the changelog entry plus a link to the Maven Central artifact;
  - the **signed jars (`*.jar`) and their detached signatures (`*.jar.asc`)** (the exact same signatures published to Maven Central, made with the shared Lambda Java release signing key, no re-signing);
  - **signature-verification instructions** in the notes (import the public key from a keyserver by fingerprint, then `gpg --verify`). Only the public key fingerprint crosses `$GITHUB_ENV`; the GPG passphrase and Sonatype token stay in-shell and are scrubbed on exit.
- New step to update a `lastPublished` marker in the module POM (see below) on each release.
- Module POMs now set a meaningful Central `deploymentName` (`groupId:artifactId:version`) so the Central Portal deployment is labeled by GAV instead of the default "DEPLOYMENT".

**`.github/workflows/release-runtime-interface-client.yml`** (RIC)

- Brought in line with `release.yml`: the required `changelogEntry` input, the shared GitHub Release step, and the `lastPublished` marker refresh. For RIC the release also attaches the native classifier jars (linux and linux_musl, x86_64 and aarch64) next to the main jar and their `.asc` signatures.
- The inline Corretto 8 setup is now the shared `setup-corretto8` action (see below).

**Shared composite actions**

- `create-github-release`: builds the notes (changelog, Central link, verify instructions) and attaches only signed, published artifacts. The asset list is derived from the `.asc` files, so unsigned intermediates like shade's `original-*.jar` never get attached. Used by both workflows.
- `setup-corretto8`: points `JAVA_HOME` and `PATH` at the runner image's preinstalled Corretto 8 (`$JAVA_8_HOME`) and writes a Maven `toolchains.xml`. Avoids `actions/setup-java`, which reaches hosts the runner egress lock blocks. Used by both workflows.

**Hardening**

- `resolve-release-version` now checks the effective release version against a strict `MAJOR.MINOR.PATCH[-prerelease]` shape and rejects `-SNAPSHOT`. That string later lands in a `sed` program, in git tag and branch names, and in `$GITHUB_ENV`, so the allowed characters deliberately leave out `sed` delimiters and metacharacters, whitespace, and newlines. This also puts release-version validation on the same footing as the existing `developmentVersion` check.
- Both workflows now pass the GPG passphrase to Maven through `MAVEN_GPG_PASSPHRASE` (the gpg plugin reads it directly) instead of `-Dgpg.passphrase=...`, keeping it out of the process argument list and out of `release:perform`'s space-split `-Darguments`.
- The RIC `release` job re-checks that it's running on `main`, since it publishes on a different runner than the `build-natives` job that already does that check.

**Module POMs** (`core`, `events`, `events-sdk-transformer`, `log4j2`, `serialization`, `tests`, `runtime-interface-client`)

- Added a `<!-- lastPublished: X -->` comment under `<version>` recording the last version published to Central, seeded with each module's current latest. Informational only (Maven ignores it); auto-updated by the release workflow.
- Added the `<deploymentName>` config to the central-publishing-maven-plugin.

**`RELEASING.md`**

- New maintainer doc: how to cut a release, what the workflow does, dry runs, and failure handling.

Note: the shared parent-POM / version-map change that was previously part of this branch has been split into a separate follow-up PR and is not included here.

*Target (OCI, Managed Runtime, both):* both


.github/workflows/release.yml
```diff
@@ -252,6 +259,62 @@ jobs:
...
+          else
+            echo "::notice::$MODULE has no lastPublished comment; skipping annotation"
+          fi
+
+          # (2) Parent version map — keyed only for modules other modules depend on.
+          case "$MODULE" in
+            aws-lambda-java-core)          PROP=lambda.core.version ;;
+            aws-lambda-java-events)        PROP=lambda.events.version ;;
+            aws-lambda-java-serialization) PROP=lambda.serialization.version ;;
+            *)                             PROP="" ;;
```
darklight3it: I don't like this hard mapping for different reasons.

- First are we assuming that we are can have multiple tags right? Is this a good experience?
- Why only those packages and not all the packages we are publishing on Maven.
- In the case we need a subset why don't you create a properties you can read from the pom that directly tells you if you need to read the version?

aws-lambda-java-core/pom.xml
```diff
@@ -3,9 +3,17 @@
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
   
+  <parent>
+    <groupId>com.amazonaws</groupId>
+    <artifactId>aws-lambda-java-libs-parent</artifactId>
```
darklight3it: I think you can just call it `aws-java-libs`, the "parentness" of that package is implied by the `<parent/>` tag.

pom.xml
```diff
@@ -0,0 +1,97 @@
...
+            <organizationUrl>https://aws.amazon.com/</organizationUrl>
+        </developer>
+    </developers>
+
+    <!-- Version map: one version per artifact. -->
+    <properties>
+        <lambda.core.version>1.4.0</lambda.core.version>
+        <lambda.events.version>3.16.1</lambda.events.version>
+        <lambda.serialization.version>1.4.1</lambda.serialization.version>
+    </properties>
```
darklight3it: Why only those packages? Also double check their version.

pom.xml
```diff
@@ -0,0 +1,97 @@
...
+                <artifactId>aws-lambda-java-events</artifactId>
+                <version>${lambda.events.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.amazonaws</groupId>
+                <artifactId>aws-lambda-java-serialization</artifactId>
+                <version>${lambda.serialization.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
darklight3it: I think in this file you can also define version number of transitive dependencies.

.github/workflows/release-runtime-interface-client.yml
```diff
@@ -319,9 +299,13 @@ jobs:
...
 
           # --- Publish --- (-gs: merge Sonatype creds with the ~/.m2 mirror)
+          # Pass the passphrase via MAVEN_GPG_PASSPHRASE (read natively by
+          # maven-gpg-plugin) rather than -Dgpg.passphrase, so it never lands in
+          # the process argument list.
+          export MAVEN_GPG_PASSPHRASE="$GPG_PASSPHRASE"
           mvn deploy -Prelease -DskipTests -DmultiArch=false \
             -gs "$MAVEN_SETTINGS" \
-            -Dgpg.keyname="$GPG_KEYNAME" -Dgpg.passphrase="$GPG_PASSPHRASE" \
+            -Dgpg.keyname="$GPG_KEYNAME" \
```
anzheyazzz, line 308: I can see the plugin `maven-gpg-plugin` version are all v1.5, does it support this new feature? Do we need to bump the version of `maven-gpg-plugin` as well?

.github/workflows/release.yml
```diff
@@ -260,18 +290,34 @@ jobs:
...
+        if: ${{ github.event.inputs.skip_publish != 'true' }}
+        uses: ./.github/actions/create-github-release
+        with:
+          tag: ${{ env.TAG_NAME }}
+          title: ${{ env.MODULE }} ${{ env.EFFECTIVE_RELEASE_VERSION }}
+          module: ${{ env.MODULE }}
+          version: ${{ env.EFFECTIVE_RELEASE_VERSION }}
+          changelog-entry: ${{ env.CHANGELOG_ENTRY_INPUT }}
+          fingerprint: ${{ env.GPG_FINGERPRINT }}
+          artifact-dir: ${{ env.MODULE }}/target/checkout/target
```
anzheyazzz, line 319: It passes `<module>/target/checkout/target` here, but I think `release:perform` builds the module under path like `<module>/target/checkout/<module>/target`

.github/workflows/release-runtime-interface-client.yml
darklight3it: Nice catch. The example provided here is with the new version which is 3.2.8

https://maven.apache.org/plugins/maven-gpg-plugin/usage.html

This seems to me something we can add later. Can we descope it in another PR?

.github/workflows/release.yml
darklight3it: +1. If I can suggest can we just create the env where mvn has created the artifact just after we create it? Also let's fail if we don't find a valid artifact.

.github/workflows/release-runtime-interface-client.yml
```diff
@@ -354,6 +355,24 @@ jobs:
             --title "chore(release): ${MODULE} ${EFFECTIVE_RELEASE_VERSION}" \
             --body "Post-release version bump for ${MODULE} ${EFFECTIVE_RELEASE_VERSION} (already on Maven Central, tag ${TAG_NAME} pushed)."
```
darklight3it, line 356: I think we can probably extract this. This is really similar code to the one in release.

PR #642, opened 2026-08-26
ci: Fix codecov project check on merge to main

fabisev:
*Issue #, if available:*

*Description of changes:*
### Problem
Some merges to main fail `codecov/project` even when the merged tree is unchanged. Last case (PR #629): "65.38% (-0.41%) compared to 7b484d3"  (a drop on a tree that is byte-for-byte identical to the PR branch tip).

### Root cause
PR builds upload coverage from two native matrix runners (x86_64 + aarch64), which Codecov merges into a union. Merge-to-main uploaded a single x86 report, so the merge commit measured slightly less coverage than its own base. With no `codecov.yml`, the default project status fails on any drop.

Codecov docs:
- https://docs.codecov.com/docs/merging-reports: "Codecov does not override report data for multiple uploads. We always merge the data."
- https://docs.codecov.com/docs/unexpected-coverage-changes, reason 2: "a different number of reports between head and base."

### Fix
**`.codecov.yml` (new):**
- `project.threshold: 1%`: tolerate sub-1% noise.
- `patch: off`: project coverage is the gate.
- `carryforward: true`: Codecov's recommended remedy for commits that don't re-upload every report (https://docs.codecov.com/docs/carryforward-flags).

**Workflows:**
- PR: tag each matrix upload with `flags: ${{ matrix.arch }}` + explicit JaCoCo path.
- Merge-to-main: add a two-arch `coverage` matrix job so main and PR record the same union of reports (also gives carryforward a real per-arch baseline).

### Result
Merges compare the union of both archs against a like-for-like base, so measurement noise no longer fails the build.

### DIFF
--- a/.github/workflows/runtime-interface-client_merge_to_main.yml
+++ b/.github/workflows/runtime-interface-client_merge_to_main.yml
@@ -89,8 +89,57 @@ jobs:
         ENABLE_SNAPSHOT: ${{ secrets.ENABLE_SNAPSHOT }}
       run: make publish
 
+  # Coverage is measured on both native architectures (x86_64 and aarch64) so
+  # the coverage recorded for a merge to main matches what the PR build recorded
+  # (the union of both archs). Uploading a single x86_64 report here previously
+  # made every merge commit look ~0.4% lower than its own PR base, failing the
+  # codecov/project status on identical code.
+  coverage:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-latest
+          - arch: aarch64
+            runner: ubuntu-24.04-arm
+    runs-on: ${{ matrix.runner }}
+    name: "coverage (${{ matrix.arch }})"
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/runtime-interface-client_merge_to_main.yml
```diff
@@ -89,8 +89,57 @@ jobs:
         ENABLE_SNAPSHOT: ${{ secrets.ENABLE_SNAPSHOT }}
       run: make publish
 
+  # Coverage is measured on both native architectures (x86_64 and aarch64) so
+  # the coverage recorded for a merge to main matches what the PR build recorded
+  # (the union of both archs). Uploading a single x86_64 report here previously
+  # made every merge commit look ~0.4% lower than its own PR base, failing the
+  # codecov/project status on identical code.
+  coverage:
```
darklight3it, line 97: Github is flagging that you should add some permission to this action.

darklight3it (approved):
I think it's fine, only add the permissions.

maxday (approved):

PR #646, opened 2026-09-02
feat: Emit worker pool size as a DEBUG log event during init on Lambda Managed Instances

vip-amzn:
## Description

On [Lambda Managed Instances](https://docs.aws.amazon.com/lambda/latest/dg/lambda-managed-instances.html),
the runtime sizes and spawns a worker pool during initialization to handle
multiple concurrent invocations per execution environment. Today there is no
way to observe the effective pool size: it can be influenced by configuration,
but nothing confirms what the runtime actually resolved.

This change adds a one-time structured JSON log event, emitted at DEBUG level
during runtime initialization, reporting the resolved worker pool size and the
maximum concurrency the execution environment supports. Both values are
included so the worker count is not mistaken for the supported concurrency.

Example (visible when the function's application log level is DEBUG or lower):

```json
{
  "timestamp": "...",
  "level": "DEBUG",
  "message": {
    "event": "runtime_worker_pool_initializing",
    "workerCount": 17,
    "executionEnvironmentMaxConcurrency": 34
  }
}
```

Design notes:
- Emitted exactly once per execution environment, during init — never per invoke.
- Opt-in by design: routed through the runtime's standard log-level filtering,
  so it only appears when the application log level is set to DEBUG or lower.
  Default configurations see no new log lines and no added CloudWatch cost.
- No new dependencies and no eager loading: emission reuses logging machinery
  already initialized during bootstrap.
- No behavior change on the standard (single-concurrency) compute type; that
  path is untouched.
- The nested message object is queryable in CloudWatch Logs Insights
  (e.g. filter message.event = "runtime_worker_pool_initializing").

See also: [Java runtime for Lambda Managed Instances](https://docs.aws.amazon.com/lambda/latest/dg/lambda-managed-instances-java-runtime.html)

## Testing

- Full unit test suite passes; new tests cover: emitted exactly once with the
  correct schema and values, suppressed at INFO and above, and not emitted on
  the single-concurrency path.
- Validated end to end on Lambda Managed Instances: event appears once per
  execution environment at DEBUG, never appears at INFO or above, and repeated
  sequential/concurrent invocations neither re-emit the event nor affect
  normal invocation logs.

### DIFF
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/logging/StructuredLogMessage.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/logging/StructuredLogMessage.java
@@ -9,7 +9,7 @@ import com.amazonaws.services.lambda.runtime.logging.LogLevel;
 
 class StructuredLogMessage {
     public String timestamp;
-    public String message;
+    public Object message;
     public LogLevel level;
     public String AWSRequestId;
     public String tenantId;
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/logging/StructuredLogMessage.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/logging/StructuredLogMessage.java
@@ -9,7 +9,7 @@ import com.amazonaws.services.lambda.runtime.logging.LogLevel;
 
 class StructuredLogMessage {
     public String timestamp;
-    public Object message;
+    public String message;
     public LogLevel level;
     public String AWSRequestId;
     public String tenantId;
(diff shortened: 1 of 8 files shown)


aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/logging/StructuredLogMessage.java
```diff
@@ -9,7 +9,7 @@
 
 class StructuredLogMessage {
     public String timestamp;
-    public String message;
```
darklight3it, line 12: This can be breaking this object is mapped through `GSON` for serialization and that could access the internals by reflection.

Besides that this is a class that is a good candidate for a customer to tweak through reflection. We should be careful here. Let's try to avoid changes.


Note: This class is unusually ugly without getter and setters, we should come to somethign better in the future.

darklight3it: Fine for me if we test.

maxday:
I'm not sure about changing the log struct from String to Object. 
Can you write a function which returns a string from a WorkerPoolInitializedEvent? Then we can keep all your existing type as is?

darklight3it (approved):

aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/logging/StructuredLogMessage.java
vip-amzn: Changed the methods to use log(bytes, level)

maxday (approved):

PR #648, opened 2026-09-11
fix: remove unused Platform serialization code path

maxday:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/EventHandlerLoader.java
+++ b/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/EventHandlerLoader.java
@@ -5,7 +5,6 @@ SPDX-License-Identifier: Apache-2.0
 
 package com.amazonaws.services.lambda.runtime.api.client;
 
-import com.amazonaws.services.lambda.runtime.ClientContext;
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.LambdaLogger;
 import com.amazonaws.services.lambda.runtime.LambdaRuntimeInternal;
@@ -38,11 +37,9 @@ import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.util.Arrays;
 import java.util.Comparator;
-import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import static com.amazonaws.services.lambda.runtime.api.client.UserFault.filterStackTrace;
 import static com.amazonaws.services.lambda.runtime.api.client.UserFault.makeUserFault;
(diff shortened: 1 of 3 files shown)


darklight3it (commented):
Hey @maxday! 👋

I agree that simplifying this logic would be valuable.

However, I found a concrete compatibility concern that makes me hesitant to classify this path as unused: the AWS SDK for Android’s Lambda invoker automatically supplies the platform value that selects it.

Specifically:

* [ClientContext](https://github.com/aws-amplify/aws-sdk-android/blob/48b1f996d0da6c5706c295e8449ccefa6c6a4517/aws-android-sdk-lambda/src/main/java/com/amazonaws/mobileconnectors/util/ClientContext.java#L145-L159) sets `env.platform` to `"Android"`.
* [LambdaInvokerFactory](https://github.com/aws-amplify/aws-sdk-android/blob/48b1f996d0da6c5706c295e8449ccefa6c6a4517/aws-android-sdk-lambda/src/main/java/com/amazonaws/mobileconnectors/lambdainvoker/LambdaInvokerFactory.java#L157-L164) constructs that client context, and [LambdaInvocationHandler](https://github.com/aws-amplify/aws-sdk-android/blob/48b1f996d0da6c5706c295e8449ccefa6c6a4517/aws-android-sdk-lambda/src/main/java/com/amazonaws/mobileconnectors/lambdainvoker/LambdaInvocationHandler.java#L83-L112) attaches it to the invocation.
* [LambdaInvokerFactory](https://github.com/aws-amplify/aws-sdk-android/blob/48b1f996d0da6c5706c295e8449ccefa6c6a4517/aws-android-sdk-lambda/src/main/java/com/amazonaws/mobileconnectors/lambdainvoker/LambdaInvokerFactory.java#L315-L316) uses [LambdaJsonBinder](https://github.com/aws-amplify/aws-sdk-android/blob/48b1f996d0da6c5706c295e8449ccefa6c6a4517/aws-android-sdk-lambda/src/main/java/com/amazonaws/mobileconnectors/lambdainvoker/LambdaJsonBinder.java#L22-L62) by default, which uses Gson for request serialization and response deserialization.

This suggests that the runtime’s Gson path was intended to provide serialization semantics compatible with the Android client. I haven’t found an explicit statement confirming that intent. The AWS SDK for Android is now archived, and I couldn’t find this invoker or an equivalent direct-invocation API in the current Amplify Android repository. Existing applications may therefore still depend on the archived SDK and this behavior.

Since this PR changes both input deserialization and output serialization, I think we should treat it as potentially breaking for those callers. If you have additional information showing that this integration has been replaced, is no longer used, or that compatibility is otherwise covered, I’m happy to revisit that assessment.

If we decide to remove this path, I see three possible approaches:

* Merge this change, remove Gson as a dependency, and release new major versions of both Serialization and the RIC from a separate branch. These versions could initially be used only by OCI-based functions and a future managed Java runtime. However, this would require maintaining the existing versions and backporting changes, so I would not recommend this approach at the moment.

* Consolidate both paths on Jackson, while configuring and testing Jackson to preserve the serialization behavior expected by Android clients. If we can establish compatibility, we could remove Gson without introducing a breaking change.

* Follow the migration path I documented last year: use a service-provider design to decouple serialization from the RIC, preserve the existing behavior by default, and allow customers to opt into a new implementation. This could also support a future Jackson major-version upgrade, although I would keep that as a separate design discussion rather than a prerequisite for this PR.

For the last option, separating these concerns internally would not necessarily require publishing every resulting module. In general, I think we should publish fewer libraries to Maven Central, each with a clear customer use case. For example, I don’t see a compelling reason for customers to consume our serialization implementation independently of the RIC.

Overall, I support the cleanup; I’d just like us to understand the compatibility impact and agree on a migration strategy before shipping it.

